PR #2, opened 2018-09-18
Use secure connection with Findologic

tomas-adeoweb:


resources/lib/http_library.php
```diff
@@ -5,9 +5,9 @@
 /** @var \Findologic\Api\Request\Request $request */
 $request = SdkRestApi::getParam('request');
 
-$httpRequest = new \HTTP_Request2();
+$method = $request['method'] ?? 'GET';
```
ihanli: I think we can safely assume `GET` for all requests since our API doesn't support anything else anyway.

PR #3, opened 2018-10-24
Send user ip with requests

tomas-adeoweb:


src/Api/Request/RequestBuilder.php
```diff
@@ -88,6 +88,18 @@ public function getCleanShopUrl($type = 'request')
         return $url;
     }
 
+    /**
+     * @return string|bool
+     */
+    public function getUserIp()
+    {
+        if (!empty($_SERVER['REMOTE_ADDR'])) {
```
ihanli: This can be rather tricky since the IP doesn't necessarily have to be at that index e.g. when a proxy is used. We can copy some code from our other plugins here.
```
        if ($_SERVER['HTTP_CLIENT_IP']) {
            $ipaddress = $_SERVER['HTTP_CLIENT_IP'];
        } elseif ($_SERVER['HTTP_X_FORWARDED_FOR']) {
            $ipaddress = $_SERVER['HTTP_X_FORWARDED_FOR'];
        } elseif ($_SERVER['HTTP_X_FORWARDED']) {
            $ipaddress = $_SERVER['HTTP_X_FORWARDED'];
        } elseif ($_SERVER['HTTP_FORWARDED_FOR']) {
            $ipaddress = $_SERVER['HTTP_FORWARDED_FOR'];
        } elseif ($_SERVER['HTTP_FORWARDED']) {
            $ipaddress = $_SERVER['HTTP_FORWARDED'];
        } elseif ($_SERVER['REMOTE_ADDR']) {
            $ipaddress = $_SERVER['REMOTE_ADDR'];
        } else {
            $ipaddress = false;
        }
```
It is important that you:
1. Check all mentioned fields.
2. Check all mentioned fields in the given order.

PR #6, opened 2018-10-24
Make Findologic's smart suggest available

tomas-adeoweb:


src/Providers/FindologicServiceProvider.php
```diff
@@ -47,6 +47,20 @@ public function boot(
             return;
         }
 
+        $logger = $this->getLoggerObject();
+
+        $eventDispatcher->listen(
+            'IO.Resources.Import',
+            function (ResourceContainer $container) use ($configRepository) {
+                $container->addScriptTemplate(
```
ihanli: Make sure that our snippet is only added if:

- [ ] Our search is active

- [ ] The shopkey is not empty

src/Providers/FindologicServiceProvider.php
```diff
@@ -47,6 +47,22 @@ public function boot(
             return;
         }
 
+        if (!$configRepository->get(Plugin::CONFIG_SHOPKEY, false)) {
```
ihanli: @tomas-adeoweb 
See [comment](https://github.com/findologic/plugin-plentymarkets-ceres-api/pull/6/files#r228881098). We still have to check if the search was activated in the plugin specific configs.

tomas-adeoweb: @ihanli The if with Plugin::CONFIG_ENABLED is checking this.

PR #7, opened 2018-11-13
Make navigation request to selector endpotnt

tomas-adeoweb:


resources/lib/tests/Api/Request/ParametersBuilderTest.php
```diff
@@ -0,0 +1,194 @@
...
+
+        foreach ($categories as $id => $category) {
+            $categoryMock = $this->getMockBuilder(Category::class)->disableOriginalConstructor()->getMock();
+            $categoryMock->parentCategoryId = $category['parentCategoryId'];
+
+            $details = new \stdClass();
+            $details->name = $category['name'];
+            $categoryMock->details = [$details];
+
+            $mockedCategories[] =[$id, $categoryMock];
```
ihanli: You are missing a white space here -> `$mockedCategories[] = [$id, $categoryMock];`

resources/lib/tests/Api/Request/ParametersBuilderTest.php
```diff
@@ -0,0 +1,194 @@
...
+
+        $result = $parametersBuilderMock->setSearchParams($requestMock, $httpRequestMock, $categoryMock);
+
+        $this->assertEquals($expectedParameters, $result->getParams());
+    }
+
+    public function providerGetCategoryName()
+    {
+        return [
+            [
```
ihanli: Add a description for this data set similar to `providerSetSearchParams`

resources/lib/tests/Api/Request/ParametersBuilderTest.php
```diff
@@ -0,0 +1,194 @@
...
+                [
+                    2 => ['parentCategoryId' => 1, 'name' => 'Test1'],
+                    1 => ['parentCategoryId' => 0, 'name' => 'Test0']
+                ],
+                'Test1_Test0'
```
ihanli: Shouldn't this be `Test0_Test1` since `Test0` seems to be the parent category?

resources/lib/tests/Api/Request/ParametersBuilderTest.php
```diff
@@ -0,0 +1,194 @@
...
+            ]
+        ];
+    }
+
+    /**
+     * @dataProvider providerGetCategoryName
```
ihanli: Add the missing PHPDoc entries.

resources/lib/tests/Api/Request/ParametersBuilderTest.php
```diff
@@ -0,0 +1,194 @@
...
+     */
+    public function testGetCategoryName($categories, $expectedCategoryName)
```
ihanli: -> `public function testGetCategoryName(array $categories, $expectedCategoryName)`

resources/lib/tests/Api/Request/ParametersBuilderTest.php
```diff
@@ -0,0 +1,194 @@
...
+     * @param array|null $methods
+     * @return ParametersBuilder|MockObject
+     */
+    protected function getParametersBuilderMock($methods = null)
+    {
+        if (!is_array($methods)) {
+            $methods = [];
+        }
+
+        $methods = array_merge($methods, ['getCategoryService']);
```
ihanli: I think it is safe to use `$methosds[] = 'getCategoryService'` instead.

src/Api/Client.php
```diff
@@ -43,6 +43,8 @@ public function call(Request $request)
         $response = false;
 
         try {
+            $this->logger->error('Request: ', $this->requestToArray($request));
```
ihanli: Remove all log entries that were added for debugging.

src/Api/Request/ParametersBuilder.php
```diff
@@ -0,0 +1,141 @@
...
+            }
+        }
+
+        if ($category && ($categoryFullName = $this->getCategoryName($category))) {
+            $request->setParam('selected', ['cat' => [$categoryFullName]]);
+            //TODO: remove after testing
+            $this->logger->error('Category full name: ' . $categoryFullName);
+        }
+
+        if (isset($parameters[Plugin::API_PARAMETER_SORT_ORDER]) && in_array($parameters[Plugin::API_PARAMETER_SORT_ORDER], Plugin::API_SORT_ORDER_AVAILABLE_OPTIONS)) {
```
ihanli: This won't work since the constant's value is `order` but Plentymarkets uses `sorting`. This was actually fixed before, seems like it happend during resolving some merge conflicts.

src/Api/Request/ParametersBuilder.php
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * @param Request $request
+     * @param array $parameters
+     * @return Request
+     */
+    public function setPagination($request, $parameters)
+    {
+        $pageSize = $parameters[Plugin::API_PARAMETER_PAGINATION_ITEMS_PER_PAGE] ?? 0;
```
ihanli: This won't work since the constant's value is `count` but Plentymarkets uses `items`. This was actually fixed before, seems like it happend during resolving some merge conflicts.

src/Api/Request/ParametersBuilder.php
```diff
@@ -0,0 +1,141 @@
...
+     * @param Category $category
+     * @return string
+     */
+    public function getCategoryName($category)
+    {
+        $categoryName = '';
+
+        try {
+            $categoryTree = $this->getCategoryTree($category);
+            array_reverse($categoryTree);
```
ihanli: This doesn't seem to fix the issue since I'm still getting the categories in the wrong order. Also the respective test implies that the wrong order is returned.

src/Services/SearchService.php
```diff
@@ -117,10 +135,16 @@ public function handleSearchOptions($searchOptions, $request)
      */
     protected function search($request)
     {
+        if ($this->results) {
```
ihanli: I guess this was added to prevent the issue of double requests:
* This could cause caching issues in the future.
* It doesn't solve the actual problem. Plenty is still trying to make the same request twice.

It would be interesting why there are double requests since it didn't before (this story).

resources/lib/tests/Api/Request/ParametersBuilderTest.php
tomas-adeoweb: I will leave this like it is for the future tests.

src/Services/SearchService.php
tomas-adeoweb: The search function is called from two methods - handleSearchOptions and handleSearchQuery. Both of them are triggered on search request but previously I only worked with handleSearchQuery. I'm not sure what could have changed but for now I only commented the code which calls search function from handleSearchOptions method. All information is parsed and formatted inside handleSearchOptions so in the future maybe it could be possible to remove handleSearchOptions if it is not needed when implementing filters.

resources/lib/tests/Api/Request/ParametersBuilderTest.php
ihanli: I had to read this 5 times before I understood it :smile: 
We can leave this as it is, since your point is valid.

PR #9, opened 2018-11-23
Send plugin revision with requests

tomas-adeoweb:


ihanli:
@tomas-adeoweb 
As discussed in our call, I will close this pull request since the code changes don't have the desired effect and have to be reverted anyway.

PR #14, opened 2019-01-08
Rename property main_variation_id

ihanli:


dblaichinger:
@ihanli 
Is this related to https://findologic.myjetbrains.com/youtrack/issue/PLENTY-50?
We should keep the way to use story ids for branch names and pull request titles.

PR #21, opened 2019-02-06
Add config option for category pages

ihanli:


ihanli:
Will be closed since this approach doesn't yield the desired result. Will be solved by #27 .

PR #27, opened 2019-02-19
Plenty 54 configuration to disable navigation integration

wladi0097:


config.json
```diff
@@ -22,5 +22,16 @@
...
+  {
+    "tab": "Findologic Search",
+    "key": "nav_enabled",
+    "label": "Navigation enabled",
+    "type": "dropdown",
+    "possibleValues": {
+      "0": "No",
+      "1": "Yes"
+    },
+    "default"   : "0"
```
ihanli: On second thought, it might be better to have this active by default.

src/Providers/FindologicServiceProvider.php
```diff
@@ -52,37 +63,47 @@ public function boot(
             return;
         }
 
+        $this->isSearchPage = strpos($request->getUri(), '/search') !== false;
+        $this->activeOnCatPage = !$this->isSearchPage && $configRepository->get(
```
ihanli: Remove the default value for `CONFIG_NAVIGATION_ENABLED` since it can never be empty.

src/Services/SearchService.php
```diff
@@ -130,8 +141,13 @@ public function doNavigation(HttpRequest $request, ExternalSearch $externalSearc
             $response->getData(Response::DATA_RESULTS)['count']
         );
 
-        $this->results->setData(Response::DATA_RESULTS, $response->getData(Response::DATA_RESULTS));
-        $this->results->setData(Response::DATA_PRODUCTS, $response->getData(Response::DATA_PRODUCTS));
+        if ($this->configRepository->get(Plugin::CONFIG_NAVIGATION_ENABLED, false)) {
```
ihanli: Remove the default value for `CONFIG_NAVIGATION_ENABLED?.

src/Services/SearchService.php
```diff
@@ -140,11 +156,10 @@ public function doNavigation(HttpRequest $request, ExternalSearch $externalSearc
     public function handleSearchQuery(HttpRequest $request, ExternalSearch $externalSearch)
     {
         try {
-            $results = $this->search($request, $externalSearch);
-            if ($externalSearch->categoryId !== null && $request->get('attrib') === null) {
+            if ($externalSearch->categoryId !== null && $request->get('attrib') === null) { // Is Navigation
```
ihanli: Put the comment on top of the if statement.

src/Services/SearchService.php
```diff
@@ -130,8 +141,13 @@ public function doNavigation(HttpRequest $request, ExternalSearch $externalSearc
             $response->getData(Response::DATA_RESULTS)['count']
```
ihanli: Try using `\Findologic\Api\Response\Response::getResultsCount` instead.

PR #29, opened 2019-02-27
PLENTY-63 remove config option enabled

darius-adeoweb:


tests/Providers/FindologicServiceProviderTest.php
```diff
@@ -1,47 +0,0 @@
-<?php
```
ihanli: Delete the whole file if it is empty now.

PR #31, opened 2019-03-05
PLENTY-64 Remove config option url

darius-adeoweb:


src/Api/Request/RequestBuilder.php
```diff
@@ -92,7 +93,7 @@ public function createRequestObject()
      */
     public function getCleanShopUrl($type = self::DEFAULT_REQUEST_TYPE)
     {
-        $url = ltrim($this->configRepository->get(Plugin::CONFIG_URL), '/') . '/';
+        $url = ltrim(self::SEARCH_SERVER_URL, '/') . '/';
```
ihanli: The `ltrim` stuff shouldn't be necessary anymore since the user can't change the URL anymore and the constant's value definitely ends with a slash.

PR #38, opened 2019-03-19
PLENTY-68 Add support for filters of type range slider

darius-adeoweb:


src/Api/Response/Parser/FiltersParser.php
```diff
@@ -35,10 +37,16 @@ public function parse($data)
...
                 }
 
-                if ($filterName === 'price') {
+                if ($filterName === 'price' && $filterData['type'] !== Plugin::FILTER_TYPE_RANGE_SLIDER) {
                     $filterData['type'] = 'price';
                 }
 
+                if ($filterData['type'] === Plugin::FILTER_TYPE_RANGE_SLIDER) {
+                    $filterData['currency'] = $filter->attributes->unit->__toString();
+                    $filterData['minPrice'] = (float)$filter->attributes->totalRange->min->__toString();
```
ihanli: Wouldn't `(float)$filter->attributes->totalRange->min` do the same?

resources/js/src/app/components/itemList/filter/ItemPriceSlider.js
```diff
@@ -0,0 +1,67 @@
...
+        this.$options.template = this.template || "#vue-item-price-slider";
+
+        const values = this.getSelectedFilterValue(this.facet.id);
+
+        this.priceFrom = values ? values.min : this.facet.minPrice;
+        this.priceTo = values ? values.max : this.facet.maxPrice;
+
+        $(function() {
+            $("#price-slider-range").slider({
+                step: 1,
```
ihanli: I think our API provides a value for this. Consider getting it from the response XML and passing it here.

PR #39, opened 2019-03-20
PLENTY-82_filters_should_work_with_empty_attrib

wladi0097:


resources/js/src/app/mixins/url.js
```diff
@@ -140,7 +140,8 @@ export default {
         updateSelectedFilters(facetId, facetValue) {
             let params = this.getSearchParams();
 
-            if (!(Constants.PARAMETER_ATTRIBUTES in params)) {
+            if (!(Constants.PARAMETER_ATTRIBUTES in params) ||
+                params[Constants.PARAMETER_ATTRIBUTES] === '') {
```
ihanli: Instead of handling the bug like this:
* Extend function `getSearchParams` so that `Constants.PARAMETER_ATTRIBUTES` isn't present in `params` if the query parameter doesn't have a value.
* Remove the `params[Constants.PARAMETER_ATTRIBUTES] === ''` checks.

PR #48, opened 2019-04-08
PLENTY-92 Implement custom color filter

darius-adeoweb:


src/Api/Services/ColorImage.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * @param string $imageUrl
+     * @return bool
+     */
+    public function isImageAccessible(string $imageUrl)
+    {
+        try {
+            $response = (string)$this->libraryCallContract->call(
```
ihanli: Rename `$response` to `$status` since it represents the return value more accurately.

src/Api/Services/ColorImage.php
```diff
@@ -0,0 +1,61 @@
...
+                ['imageUrl' => $imageUrl]
+            );
+        } catch (\Exception $e) {
+            $this->logger->error('Exception while accessing image.');
+            $this->logger->logException($e);
+
+            return false;
+        }
+
+        if ($response[0] !== '2') {
```
ihanli: Since 3xx would also count as successful, check if the status code start with 4 or 5.

PR #57, opened 2019-04-25
PLENTY-91 Implement vendor logos

darius-adeoweb:


src/Api/Response/Parser/FiltersParser.php
```diff
@@ -107,6 +108,16 @@ public function parseFilterItem($filterType, &$filterItem, $data, $index)
                 $filterItem['priceMax'] = $data->parameters->max;
             }
 
+            if ($filterType === Plugin::FILTER_TYPE_IMAGE) {
+                if (isset($data->image) && $data->image->__toString() !== '' && $data->image->__toString()[0] !== '/') {
+                    $filterItem['imageUrl'] = null;
+
+                    if ($this->colorImageService->isImageAccessible($data->image->__toString())) {
```
ihanli: Since it is not only about colors anymore:
* Rename `$this->colorImageService` to `$this->imageService`.
* Also rename the class itself accordingly.

tests/Api/Response/Parser/FiltersParserTest.php
```diff
@@ -0,0 +1,299 @@
...
+
+        return $filtersParserMock->getMock();
+    }
+
+    /**
+     * @return \SimpleXMLElement
+     */
+    protected function getNoFiltersResponse()
+    {
+        return simplexml_load_string(preg_replace('/&(?!#?[a-z0-9]+;)/', '&amp;','<?xml version="1.0"?>
```
ihanli: Why is the `preg_replace` necessary?

tests/Api/Response/Parser/FiltersParserTest.php
```diff
@@ -0,0 +1,299 @@
...
+
+    public function testParse()
+    {
+        $responseMock = $this->getMockBuilder(Response::class)->disableOriginalConstructor()->setMethods(null)->getMock();
+        /** @var FiltersParser|MockObject $filtersParserMock */
+        $filtersParserMock = $this->getFiltersParserMock(['createResponseObject']);
+        $filtersParserMock->expects($this->any())->method('createResponseObject')->willReturn($responseMock);
+
+        $results = $filtersParserMock->parse($this->getNoFiltersResponse());
+        $this->assertEmpty($results);
```
ihanli: I would propose that you either extract the case with no filters in a separate test or you use a data provider to make it clear that this test covers multiple cases.

tests/Api/Response/Parser/FiltersParserTest.php
```diff
@@ -0,0 +1,299 @@
...
+
+        $response = $this->getResponse();
+        $results = $filtersParserMock->parse($response);
+
+        $this->assertNotEmpty($results);
+
+        foreach ($results as $index => $result) {
+            $this->assertEquals($result['id'], $response->filters->filter[$index]->name);
```
ihanli: Also add assertions which check if the array key is actually available.

tests/Api/Response/Parser/FiltersParserTest.php
darius-adeoweb: I have removed _preg_replace_ from _getNoFiltersResponse()_, since there are no ampersands. There are ampersands in _getResponse()_, and it is not parsed by _simplexml_load_string_, so needed to replace it with &amp; .

darius-adeoweb: I have added preg_replace to replace ampersand symbol with html friendly ampersand. But now I have removed preg_replace and changed the values itself in xml feed, since it is testing data, so I guess it will be clearer.

tests/Api/Response/Parser/FiltersParserTest.php
```diff
@@ -0,0 +1,314 @@
...
+            $this->assertArrayHasKey('name', $result);
+            $this->assertArrayHasKey('select', $result);
+            $this->assertArrayHasKey('type', $result);
+            $this->assertArrayHasKey('values', $result);
+
+            $this->assertEquals($result['id'], $response->filters->filter[$index]->name);
+            $this->assertEquals($result['name'], $response->filters->filter[$index]->display);
+            $this->assertEquals($result['select'], $response->filters->filter[$index]->select);
+            $this->assertEquals($result['type'], $response->filters->filter[$index]->type);
+            $this->assertEquals(count($result['values']), count($response->filters->filter[$index]->items->item));
```
ihanli: There should be an assertion method available explicitly for this case. I think it is called something like `assertCount`. Try to use that instead of `assertEquals`.

tests/Api/Response/Parser/FiltersParserTest.php
```diff
@@ -0,0 +1,314 @@
...
+
+        $results = $filtersParserMock->parse($this->getNoFiltersResponse());
+        $this->assertEmpty($results);
+    }
+
+    /**
+     * @param array|null $methods
+     * @return FiltersParser|MockObject
+     */
+    protected function getFiltersParserMock($methods = null)
```
ihanli: Wouldn't an empty array more appropriate as the default value?

PR #59, opened 2019-05-14
PLENTY-113 Negative alive test does not break shop pages

darius-adeoweb:


src/Containers/SearchBarContainer.php
```diff
@@ -12,10 +13,17 @@ class SearchBarContainer
 {
     /**
      * @param Twig $twig
+     * @param SearchService $searchService
      * @return string
      */
-    public function call(Twig $twig):string
+    public function call(Twig $twig, SearchService $searchService):string
     {
+        if (!$searchService->isAliveTest()) {
```
ihanli: I click tested this real quick cause it would be strange that this doesn't work without an alive test since this template doesn't need any data from our API.
Turns out that this alive check is indeed not necessary. Please apply all changes mentioned in this review and check if you can confirm my click test result.

src/Providers/FindologicServiceProvider.php
```diff
@@ -59,6 +59,10 @@ public function boot(
             return;
         }
 
+        if (!$searchService->isAliveTest()) {
```
ihanli: Using the implementation as described in the story should work now since the alive test will not be performed in any other place.

tests/Providers/FindologicServiceProviderTest.php
```diff
@@ -0,0 +1,129 @@
...
+        $this->eventDispatcher->expects($this->at(3))->method('listen')->with('Ceres.Search.Query');
+
+        $this->request->method('getUri')->willReturn('https://testshop.com/testpage');
+
+        $this->runBoot();
+    }
+
+    protected function runBoot()
+    {
+        (new FindologicServiceProvider())->boot(
```
ihanli: Assign `new FindologicServiceProvider()` to a local variable, just so it is more readable.

tests/Providers/FindologicServiceProviderTest.php
```diff
@@ -0,0 +1,129 @@
...
+        $this->request = $this->getMockBuilder(Request::class)->disableOriginalConstructor()->setMethods([])->getMock();
+        $this->eventDispatcher = $this->getMockBuilder(Dispatcher::class)->disableOriginalConstructor()->setMethods([])->getMock();
+        $this->configRepository = $this->getMockBuilder(ConfigRepository::class)->disableOriginalConstructor()->setMethods([])->getMock();
+    }
+
+    public function testBootShopKeyNotSet()
+    {
+        $this->configRepository->expects($this->once())->method('get')->willReturn('');
+
+        $this->eventDispatcher->expects($this->never())->method('listen');
```
ihanli: Add a check which makes sure that the alive test wasn't triggered.

tests/Providers/FindologicServiceProviderTest.php
```diff
@@ -0,0 +1,129 @@
...
+
+        $this->runBoot();
+    }
+
+    public function testBootShopKeySetAndAliveTestFails()
+    {
+        $this->configRepository->expects($this->once())->method('get')->willReturn('testShopKey');
```
ihanli: Explicitly mock the alive test to return something invalid and add a check that it was called once.

tests/Providers/FindologicServiceProviderTest.php
```diff
@@ -0,0 +1,129 @@
...
+    {
+        $this->configRepository->expects($this->any())->method('get')->willReturnMap([
+            [
+                Plugin::CONFIG_SHOPKEY, false, 'testShopKey'
+            ]
+        ]);
+
+        $this->searchService->expects($this->once())->method('isAliveTest')->willReturn(true);
+
+        $this->eventDispatcher->expects($this->at(0))->method('listen')->with('IO.Resources.Import');
```
ihanli: Try to use `withConsecutive` instead of providing a call index, as this can give us a hard time in the future.

tests/Providers/FindologicServiceProviderTest.php
```diff
@@ -0,0 +1,129 @@
...
```
ihanli: Try to use `withConsecutive` instead of providing a call index, as this can give us a hard time in the future.

tests/Providers/FindologicServiceProviderTest.php
```diff
@@ -0,0 +1,129 @@
...
+        $this->runBoot();
+    }
+
+    public function testIsNotSearchPageAndIsActiveOnCatPage()
+    {
+        $this->configRepository->expects($this->any())->method('get')->willReturn('testConfigValue');
+
+        $this->searchService->expects($this->once())->method('isAliveTest')->willReturn(true);
+
+        $this->eventDispatcher->expects($this->at(0))->method('listen')->with('IO.Resources.Import');
```
ihanli: Try to use `withConsecutive` instead of providing a call index, as this can give us a hard time in the future.

src/Containers/SearchFilterContainer.php
```diff
@@ -14,6 +14,12 @@ class SearchFilterContainer
 {
     public function call(Twig $twig, SearchService $searchService):string
     {
+        if (!$searchService->isAliveTest()) {
```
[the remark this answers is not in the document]
darius-adeoweb: I have found that the Internal Server Error was thrown, because in _SearchFilterContainer.php_, _$searchResults_ was null. 
So removed the check in this class, and added check for null in _SearchFilterContainer.php_.

tests/Providers/FindologicServiceProviderTest.php
darius-adeoweb: I have changed to that it mocks throwing exception, because in _FindologicServiceProvider.php_, it is catching the exception and returning, not checking the return value of _aliveTest()_

PR #68, opened 2019-06-20
PLENTY-117 Support multi language shops

darius-adeoweb:


darius-adeoweb:
Fixed travis build. It started using PHPUnit 8, which is not compatible with PHP 7.1. So changed that it would use PHPUnit installed by Composer.

PR #69, opened 2019-06-20
PLENTY-118 Update fstream dependency

darius-adeoweb:


darius-adeoweb:
Fixed travis build. It started using PHPUnit 8, which is not compatible with PHP 7.1. So changed that it would use PHPUnit installed by Composer.

.travis.yml
```diff
@@ -13,4 +13,4 @@ install:
   - php composer.phar install
 
 script:
-  - phpunit -c tests/phpunit.xml
\ No newline at end of file
+  - php vendor/phpunit/phpunit/phpunit -c tests/phpunit.xml
```
ihanli: Try using `vendor/bin/phpunit` instead. Should actually be the same file but the path is more readable.

PR #70, opened 2019-06-27
PLENTY-95 Add support for more sorting options

darius-adeoweb:


tests/Services/Search/ParametersHandlerTest.php
```diff
@@ -0,0 +1,318 @@
...
+use PHPUnit\Framework\MockObject\MockObject;
+use Plenty\Plugin\Http\Request;
+
+/**
+ * Class SearchServiceTest
+ * @package Findologic\Tests
+ */
+class ParametersHandlerTest extends TestCase
+{
+    public function providerGetSortingOptions()
```
ihanli: We should use the same naming convention for data provider as we do everywhere: type of returned data + "Provider" e.g. `sortingOptionsProvider`

tests/Services/Search/ParametersHandlerTest.php
```diff
@@ -0,0 +1,318 @@
...
+    {
+        return [
+            '$isSearch is false' => [
```
ihanli: Data set descriptions should be in "clear text" since you want to describe scenarios rather than technical details:
* Sorting options for category page
* Sorting options for search page
* Force option "Relevance" for search page

tests/Services/Search/ParametersHandlerTest.php
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    /**
+     * @dataProvider providerGetSortingOptions
+     *
+     * @param bool $isSearch
+     * @param array $ceresConfigData
+     * @param array $expectedResult
+     */
+    public function testGetSortingOptions($isSearch, $ceresConfigData, $expectedResult)
```
ihanli: Add the type hints for `$ceresConfigData` and `$expectedResult`.

tests/Services/Search/ParametersHandlerTest.php
```diff
@@ -0,0 +1,318 @@
...
+    {
+        $config = $this->createCeresConfigObject($ceresConfigData);
+
+        /** @var ParametersHandler|MockObject $parametersHandlerMock */
+        $parametersHandlerMock = $this->getMockBuilder(ParametersHandler::class)
+            ->setMethods(['getConfig'])
+            ->getMock();
+        $parametersHandlerMock->expects($this->once())->method('getConfig')->willReturn($config);
+
+        $this->assertEquals($parametersHandlerMock->getSortingOptions($isSearch), $expectedResult, true);
```
ihanli: What does the parameter `true` do? Shouldn't that actually be an assertion message?

tests/Services/Search/ParametersHandlerTest.php
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    public function providerHandlePaginationAndSorting()
```
ihanli: -> `paginationAndSortingProvider`

tests/Services/Search/ParametersHandlerTest.php
```diff
@@ -0,0 +1,318 @@
...
+    {
+        return [
+            '$isSearch is true and defaultSortingSearch is texts.name1_asc' => [
```
ihanli: Use these scenario descriptions:
* Sorting options for search page with supported default
* Sorting options for category page with supported default
* Sorting options for category page with unsupported default

tests/Services/Search/ParametersHandlerTest.php
```diff
@@ -0,0 +1,318 @@
...
+                ],
+                'item.score'
+            ]
+        ];
+    }
+
+    /**
+     * @dataProvider providerHandlePaginationAndSorting
+     *
+     * @param $isSearch
```
ihanli: Add the type hints.

tests/Services/Search/ParametersHandlerTest.php
```diff
@@ -0,0 +1,318 @@
...
+     * @param $defaultSortingSearch
+     * @param $configData
+     * @param $sortingOptions
+     * @param $defaultOption
+     */
+    public function testHandlePaginationAndSorting($isSearch, $defaultSortingSearch, $configData, $sortingOptions, $defaultOption)
```
ihanli: Add the type hints.

tests/Services/Search/ParametersHandlerTest.php
```diff
@@ -0,0 +1,318 @@
...
+    {
+        /** @var ExternalSearchOptions|MockObject $externalSearchOptionsMock */
+        $externalSearchOptionsMock = $this->getMockBuilder(ExternalSearchOptions::class)
```
ihanli: Looks like we could use the actual implementation instead of a mock?

tests/Services/Search/ParametersHandlerTest.php
```diff
@@ -0,0 +1,318 @@
...
+            ->setMethods([])
+            ->getMock();
+
+        if ($isSearch) {
+            $requestMock->expects($this->once())->method('getUri')->willReturn('test.com/search');
+        }
+
+        $config = $this->createCeresConfigObject(
+            $configData,
+            $defaultSortingSearch,
```
ihanli: Rename this to `$defaultSorting` since it could also contain the sorting for category pages.

tests/Services/Search/ParametersHandlerTest.php
```diff
@@ -0,0 +1,318 @@
...
+        $this->assertEquals($externalSearchOptionsMock->getDefaultSortingOption(), $defaultOption);
+    }
+
+    /**
+     * @param array $sortingData
+     * @param string $defaultSortingSearch
+     * @param string $defaultSorting
+     * @return object
+     */
+    protected function createCeresConfigObject($sortingData = [], $defaultSortingSearch = '', $defaultSorting = '')
```
ihanli: Add type hint for `$sortingData`.

tests/Services/Search/ParametersHandlerTest.php
```diff
@@ -0,0 +1,318 @@
...
+    {
+        return $config = (object) array(
+            'sorting' => (object) array(
+                'data' => $sortingData,
+                'defaultSortingSearch' => $defaultSortingSearch,
+                'defaultSorting' => $defaultSorting
+            )
+        );
+    }
+}
```
ihanli: Add a newline at the end of the file.

tests/Services/Search/ParametersHandlerTest.php
darius-adeoweb: But in other files naming convention is 'provider' + 'name of the test'. Should I rename in other places as well?

ihanli: That one must have slipped through. Please change both occurrences.

tests/Services/Search/ParametersHandlerTest.php
darius-adeoweb: I have removed it.

PR #77, opened 2019-07-08
PLENTY-110 Get current plugin version from system

darius-adeoweb:


src/Api/Request/RequestBuilder.php
```diff
@@ -138,7 +140,18 @@ public function getUserIp()
      */
     public function getPluginVersion()
     {
-        return Plugin::PLUGIN_VERSION;
+        /** @var PluginRepositoryContract $pluginRepository */
+        $pluginRepository = pluginApp(PluginRepositoryContract::class);
+
+        /** @var PluginModel $findologicPlugin */
+        $findologicPlugin = $pluginRepository->searchPlugins(['name' => Plugin::PLUGIN_NAMESPACE])->getResult()[0];
```
ihanli: Before accessing the first element: Check if `::getResult` returns an empty array and if so, return `Plugin::PLUGIN_VERSION_UNKNOWN`. Just as an safety measure.

PR #81, opened 2019-07-11
Plenty 129 add support for filter type dropdown

darius-adeoweb:


resources/js/src/app/components/itemList/filter/ItemDropdown.js
```diff
@@ -0,0 +1,72 @@
...
+
+        onClick(event) {
+            if (!this.$el.children[0].contains(event.target)) {
+                this.hideDropdown();
+            } else {
+                this.showDropdown();
+            }
+        },
+
+        selected: function (value) {
```
darius-adeoweb: This doesn't work. It doesn't remove the current filter and set the required filter correctly.

resources/js/src/app/mixins/url.js
```diff
@@ -155,7 +155,7 @@ export default {
                     min: facetValue.min,
                     max: facetValue.max
                 };
-            } else if (this.facet.select === 'single' && facetId !== 'cat') {
+            } else if ((this.facet.select === 'single' && facetId !== 'cat') || (this.facet.type === 'select' && this.facet.select === 'multiple')) {
```
darius-adeoweb: Thought about adding the case for dropdown here. Only not sure because of select=multiple . Seems like it should be multiple selection not single.

resources/js/src/app/components/itemList/filter/ItemDropdown.js
```diff
@@ -0,0 +1,54 @@
...
+
+        showDropdown: function () {
+            this.isShowDropdown = true;
+        },
+
+        hideDropdown: function () {
+            this.isShowDropdown = false;
+        }
+    }
+});
```
ihanli: Add a newline at the end of the file.

resources/js/src/app/components/itemList/filter/ItemDropdown.js
```diff
@@ -0,0 +1,54 @@
...
+        }
+    },
+
+    created() {
+        this.$options.template = this.template || "#vue-item-dropdown";
+
+        window.addEventListener('click', this.onClick);
+    },
+
+    beforeDestroy() {
```
ihanli: Shouldn't this rather remove the event listener than adding it?

resources/js/src/app/components/itemList/filter/ItemDropdown.js
```diff
@@ -0,0 +1,54 @@
...
+    data: function () {
+        return {
+            isShowDropdown: false
+        }
+    },
+
+    created() {
+        this.$options.template = this.template || "#vue-item-dropdown";
+
+        window.addEventListener('click', this.onClick);
```
ihanli: Please check if the event listener can be set via Vue.js instead of pure JS (like it's done with `selected`).

darius-adeoweb: Using Vue.js @click, I can listen for the event when it is clicked only on the component. In this case I need to check for any click on the window, so that I know is it clicked inside the component or the outside (and then I show or hide the component respectively). 
I don't know how to set this kind of event differently, using Vue.js components

ihanli: What if you used the click event on the component to open it and the blur event to close it again once it loses focus?

darius-adeoweb: Tried it. Blur event doesn't fire.

PR #86, opened 2019-07-23
PLENTY-133 Support landing pages

darius-adeoweb:


tests/Api/Response/ResponseParserTest.php
```diff
@@ -39,12 +39,19 @@ public function setUp()
         $this->loggerFactory->expects($this->any())->method('getLogger')->willReturn($this->logger);
     }
 
+    /**
+     * @runInSeparateProcess
```
ihanli: Why is this necessary?

src/Api/Response/ResponseParser.php
```diff
@@ -41,9 +41,9 @@ public function parse($responseData)
 
         try {
             $data = $this->loadXml($responseData);
+            $this->handleLandingPage($data);
```
ihanli: For better testability I would suggest:
* Call `parseLandingPage`.
* Pass the return value (if valid) to `handleLandingPage`.
* Instead of mocking `parseLandingPage` in the test, check if `handleLandingPage` has been called with the correct parameter.

src/Api/Response/ResponseParser.php
```diff
@@ -123,7 +134,7 @@ protected function parseLandingPage(\SimpleXMLElement $data)
     {
         $landingPage = [];
 
-        if (!empty($data->landingPage) ) {
+        if (isset($data->landingPage) && !empty($data->landingPage->attributes()) ) {
```
ihanli: Since there is no apparent reason to return an array (anymore): Return the link as string or `null` if no link was provided.

tests/Api/Response/ResponseParserTest.php
```diff
@@ -39,12 +39,19 @@ public function setUp()
...
+     */
     public function testParse()
     {
         $responseMock = $this->getMockBuilder(Response::class)->disableOriginalConstructor()->setMethods(null)->getMock();
         /** @var ResponseParser|MockObject $responseParserMock */
-        $responseParserMock = $this->getResponseParserMock(['createResponseObject']);
+        $responseParserMock = $this->getResponseParserMock(['createResponseObject', 'parseLandingPage']);
         $responseParserMock->expects($this->any())->method('createResponseObject')->willReturn($responseMock);
+        $responseParserMock->expects($this->once())
```
ihanli: Check my previous comment regarding testability.

tests/Api/Response/ResponseParserTest.php
darius-adeoweb: It fails the test without it. The problem is with location header being added. 
Found a fix with process isolation here: https://github.com/sebastianbergmann/phpunit/issues/720

src/Api/Response/ResponseParser.php
ihanli: This method still returns an array when it actually should return the string representation of `$data->landingPage->attributes()->link` or `null`. Don't forget to update the tests and PHPDocs.

PR #87, opened 2019-07-24
PLENTY-134 Support promotions

darius-adeoweb:


resources/css/findologic.css
```diff
@@ -91,4 +91,11 @@
 .fl-item-color-tiles-container .fl-item-color-tiles-list .fl-item-color-tiles-list-item .fl-color-tile-label .fl-color-tile-selected-image {
     position: absolute;
     z-index: 2;
+}
+
+#fl-promotion {
+    text-align: center;
+}
+#fl-promotion .image{
```
ihanli: There should be a single whitespace between `.image` and `{`.

src/Containers/PromotionContainer.php
```diff
@@ -0,0 +1,35 @@
...
+        $searchResults = $searchService->getResults();
+
+        return $twig->render(
+            'Findologic::Category.Item.Partials.Promotion',
+            [
+                'promotion' => $searchResults->getData(Response::DATA_PROMOTION)
+            ]
+        );
+    }
+}
```
ihanli: Add a newline at the end of this file.

plugin.json
```diff
@@ -40,9 +40,23 @@
...
     }
   ],
   "dependencies": {
     "findologic/http_request2": "2.3.1"
-  }
+  },
+  "containers": [
+    {
+      "key": "CategoryItem.Promotion",
+      "name": "Category item list: Main container",
```
ihanli: Maybe we should change `Main container` to `Add content to main container`. The current name doesn't give you much info about what the container actually does.

PR #88, opened 2019-07-24
PLENTY-135 Update lodash dependency

darius-adeoweb:


darius-adeoweb:
Created pull request.
I have updated gulp to v4. That updated lodash dependency. Also ran npm audit fix to fix other high/moderate vulnerabilities found in the packages.
Current gulpfile.js was not working with v4, so updated it.

For testing added a console.log to make sure that gulp works correctly (screenshot in attachments).

PR #94, opened 2019-08-05
PLENTY-143 Use scss

darius-adeoweb:


resources/views/content/styles.twig
```diff
@@ -1,2 +1,2 @@
 <link rel="stylesheet" href="https://code.jquery.com/ui/1.12.1/themes/smoothness/jquery-ui.css" />
-<link rel="stylesheet" href="{{ plugin_path("Findologic") }}/css/findologic.css" />
\ No newline at end of file
+<link rel="stylesheet" href="{{ plugin_path("Findologic") }}/css/findologic.min.css" />
```
ihanli: As described in the story, `findologic.min.css` is supposed to be loaded additionally, not instead of `findologic.css`.

resources/scss/findologic.scss
```diff
@@ -0,0 +1,97 @@
+.findologic-container {
```
ihanli: Delete this SCSS file and keep the CSS version of it. There is a separate story for converting everything.

PR #97, opened 2019-08-13
PLENTY-125 Implement smart did you mean

darius-adeoweb:


src/Api/Response/ResponseParser.php
```diff
@@ -198,4 +200,26 @@ protected function parseProducts(\SimpleXMLElement $data)
...
         return $products;
     }
+
+    /**
+     * @param SimpleXMLElement $data
+     * @return array|null
+     */
+    protected function parseSmartDidYouMean(SimpleXMLElement $data)
+    {
+        $originalQuery = $data->query->originalQuery;
```
ihanli: Cast all these values into strings. I'm shocked that this doesn't produce any errors since it returns the XML element, not the actual string value.
Also update the corresponding test, since you will now get strings instead of `SimpleXMLElement`.

tests/Api/Response/ResponseParserTest.php
```diff
@@ -62,6 +63,122 @@ public function testParse()
...
+                        <limit first="0" count="10"/>
+                        <queryString>Test</queryString>
+                        <searchedWordCount>1</searchedWordCount>
+                        <foundWordCount>1</foundWordCount>
+                    </query>
+                    <results><count>0</count></results>
+                    <products/>
+                    <filters/>
+                </searchResult>'
+                ,
```
ihanli: This doesn't have to have its own line :wink: 
Put it on the same line as `</searchResult>'` (like it's done for the other scenarios).

src/Api/Response/ResponseParser.php
darius-adeoweb: I think it does not produce any errors, because it automatically casts the value to string when it is used. 
I have updated the pull request based on the comment. Also retested it with the cases as in the screenshots in Youtrack story.

PR #102, opened 2019-08-28
Plenty 144 get other filters from api response 2

darius-adeoweb:


resources/views/ItemList/Components/Filter/ItemFilterList.twig
```diff
@@ -1,14 +1,26 @@
 <script type="x/template" id="vue-item-filter-list">
-    <div class="findologic-filter-wrapper" v-show="facets && facets.length > 0">
```
ihanli: Shouldn't we keep the `v-show` part?

src/Api/Response/ResponseParser.php
```diff
@@ -198,4 +198,22 @@ protected function parseProducts(\SimpleXMLElement $data)
...
+     * @param \SimpleXMLElement $data
+     * @return array
+     */
+    protected function getFilters($data)
+    {
+        $mainFilters = $this->filtersParser->parse($data->filters->main);
+        $otherFilters = $this->filtersParser->parse($data->filters->other);
+
+        if (!empty($mainFilters)) {
+            foreach ($mainFilters as $mainFilterKey => $mainFilterValue) {
```
ihanli: I don't like the idea that we have to iterate over all main filters again. How about rewriting `\Findologic\Api\Response\Parser\FiltersParser::parse` so that we can add the `isMain` flag there?
In `ResponseParser::parse` you could then do `$response->setData(Response::DATA_FILTERS, $this->filtersParser->parse($data->filters));`.

resources/scss/views/ItemList/Components/Filter/item-filter-list.scss
```diff
@@ -1,8 +1,97 @@
...
+        margin-left: -15px;
+        padding-right: 15px;
+        top: 100%;
 
-.list-controls #filterCollapse .page-content .card .form-check .filter-badge {
-    top: 1.5em;
+        .page-content {
+            background: white;
+            padding: 2em !important;
+            border: 1px solid #F4F4F4;
```
ihanli: Aren't there any predefined variables for all these colors?

darius-adeoweb: I have made that there is one main Findologic.scss helper file that would import all other needed scss files. And Findologic is imported inside item-filter-list.scss.

I think a better solution would be to have one main Findologic.scss file that would import all needed scss files from other modules (Ceres) and import local scss (item-filter-list.scss). 
But for this we would need to rewrite gulp task, because right now it compiles all scss files found in resources/scss/ folder

resources/scss/Findologic.scss
```diff
@@ -0,0 +1 @@
+@import '../../vendor/plentymarkets/plugin-ceres/resources/scss/variables';
```
ihanli: * Rename this file to `findologic.scss`.
* Import also `resources/scss/views/ItemList/Components/Filter/item-filter-list.scss` and change to gulp task so it only compiles `findologic.scss`.

resources/scss/views/ItemList/Components/Filter/item-filter-list.scss
```diff
@@ -1,8 +1,99 @@
-.findologic-filter-wrapper #filterCollapse {
-    position: relative;
-    z-index: 0;
-}
+@import '../../../../Findologic';
```
ihanli: This line will have to be removed once everything is imported in `findologic.scss`.

src/Api/Response/Parser/FiltersParser.php
```diff
@@ -141,4 +116,51 @@ public function parseFilterItem($filterType, &$filterItem, $data, $index)
             }
         }
     }
+
+    /**
+     * @param $filter
```
ihanli: Add the typehint for `$filter`.

src/Api/Response/Parser/FiltersParser.php
```diff
@@ -141,4 +116,51 @@ public function parseFilterItem($filterType, &$filterItem, $data, $index)
...
+        $filterName = $filter->name->__toString();
+        $filterData = [
+            'id' => $filterName,
+            'name' => $filter->display->__toString(),
+            'select' => $filter->select->__toString(),
+            'type' => ''
+        ];
+
+        if ($isMainFilter) {
+            $filterData['isMain'] = true;
```
ihanli: Why not just add this flag even if it is false?

PR #124, opened 2019-11-12
PLENTY-167 Range slider filter must work for plentymarkets ceres

darius-adeoweb:


darius-adeoweb:
For minified CSS: 
I am guessing that there were some changes done to CSS on development branch, but gulp was not used to create a minified version.

resources/js/src/app/mixins/url.js
```diff
@@ -257,6 +257,14 @@ export default {
             return selectedFilters;
         },
 
+        /**
+         * @param attributeValue
+         * @returns {boolean}
+         */
+        isRangeSliderFilter(attributeValue) {
+            return (typeof attributeValue.min !== 'undefined' && typeof attributeValue.max)
```
TheKeymaster: I think you forgot something in the second condition? It is just `typeof attributeValue.max` but you are not checking what it should be :smile:.

darius-adeoweb: :man_facepalming:

PR #125, opened 2019-11-18
PLENTY-169 When using shopbuilder checkout there should be no vue errors

darius-adeoweb:


src/Middlewares/Middleware.php
```diff
@@ -135,6 +149,20 @@ public function after(Request $request, Response $response): Response
         return $response;
     }
 
+    /**
+     * @param string $uri
+     * @return bool
+     */
+    protected function isBlacklistedUri($uri) {
```
TheKeymaster: Add missing typehints. We won't need the docblock then.

PR #126, opened 2019-11-18
PLENTY-170 Three filterboxes in each row in firefox

darius-adeoweb:


resources/scss/views/ItemList/Components/Filter/item-filter-list.scss
```diff
@@ -1,4 +1,7 @@
 .list-controls {
+    .card-columns {
```
TheKeymaster: Interesting... How about the solution in the story? Didn't it work?

darius-adeoweb: Using "card col" classes instead of "card col-md-4" makes all filters to try to fit in one row instead of three per row.

PR #129, opened 2019-11-26
PLENTY-174_fix_spaces

TheKeymaster:
I took this opportunity to also update the vendors:

* Plenty Ceres Plugin
* IO Plugin


composer.json
```diff
@@ -4,18 +4,27 @@
   "description": "Findologic api plugin for plentymarkets system",
   "require": {
     "php": "^7.0",
-    "findologic/http_request2": "2.3.1"
+    "findologic/http_request2": "2.3.1",
+    "plentymarkets/plugin-ceres": "4.4.2",
+    "plentymarkets/plugin-io": "4.4.1"
```
dblaichinger: Add back to `require-dev`

resources/js/dist/filters-component.js
```diff
@@ -112,7 +112,8 @@ Vue.component("item-search", {
 
             _this.$store.commit("setItemListSearchString", urlParams.query);
 
-            _this.$refs.searchInput.value = urlParams.query ? urlParams.query : '';
+            var rawQuery = urlParams.query ? urlParams.query : '';
+            _this.$refs.searchInput.value = decodeURIComponent(rawQuery.replace(/\+/g, ' '));
```
dblaichinger: Add comment why we explicitly replace the `+`

resources/js/src/app/components/itemList/ItemSearch.js
```diff
@@ -50,7 +50,8 @@ Vue.component("item-search", {
 
             this.$store.commit("setItemListSearchString", urlParams.query);
 
-            this.$refs.searchInput.value = urlParams.query ? urlParams.query : '';
+            let rawQuery = urlParams.query ? urlParams.query : '';
+            this.$refs.searchInput.value = decodeURIComponent(rawQuery.replace(/\+/g, ' '));
```
dblaichinger: Add comment why we explicitly replace the `+`

resources/js/src/app/components/itemList/ItemSearch.js
```diff
@@ -50,7 +50,8 @@ Vue.component("item-search", {
 
             this.$store.commit("setItemListSearchString", urlParams.query);
 
-            this.$refs.searchInput.value = urlParams.query ? urlParams.query : '';
+            let rawQuery = urlParams.query ? urlParams.query : '';
```
dblaichinger: In `resources/js/dist/filters-component.js` you used `var` statement, we should be consistent.

TheKeymaster: That's because `filters-component.js` is the compiled one and should work for all browsers. I didn't wrote that, `gulp` compiled it for me.

PR #132, opened 2019-12-04
Removed loading of findologic.css, the minified version is enough

dblaichinger:


TheKeymaster:
@dblaichinger removing this PR, as you've created two PRs for the same branch.

PR #137, opened 2019-12-17
PLENTY-184_restore_multi_language_compatibility

TheKeymaster:


resources/views/ItemList/Components/ItemSearch.twig
```diff
@@ -1,5 +1,5 @@
 <script type="x/template" id="vue-findologic-item-search">
-    <form method="GET" action="/search">
+    <form method="GET" action="/search" onsubmit="return onFlFormSubmit(event)">
```
dblaichinger: Isn't it possible to make a simple `onsubmit="return false;"` here?

TheKeymaster: true, that works as well. Less overhead and much cleaner. Thanks for your suggestion.

PR #142, opened 2020-01-27
PLENTY-185 Show suitable search result from smart suggest

darius-adeoweb:


src/Api/Response/Response.php
```diff
@@ -90,4 +101,153 @@ public function getResultsCount()
...
+        }
+    }
+
+    private function getSelectedCategory(): array
+    {
+        if (empty($filters = $this->getData(Response::DATA_FILTERS))) {
+            return [];
+        }
+
+        foreach ($filters as $filter) {
```
TheKeymaster: I think the solution is well thought, but a bit too complex. You have like a loop inside of a loop which is inside of another loop. If for some reason there would be a bug in this, it would be incredibly hard to debug.
I'd suggest picking the category and vendor not out of the response, but out of the request. This way you will only need one loop for fetching category and vendor at the same time.

E.g.

```php
foreach ($queryParams as $key => $queryParam) {
    switch ($key) {
        case 'cat':
            // We have a category; Yay!
        case 'vendor':
            // Manufacturer it is!
        default:
            continue;
    }
}
```

src/Api/Response/ResponseParser.php
```diff
@@ -201,27 +205,60 @@ protected function parseProducts(SimpleXMLElement $data)
...
+    private function getSelectedCategoryName(array $requestParams)
+    {
+        if (!$selectedCategory = isset($requestParams['attrib']['cat']) ? $requestParams['attrib']['cat'][0] : null) {
+            return null;
+        }
+
+        if (strpos($selectedCategory, '_') !== false) {
+            $categories = explode('_', $selectedCategory);
+
+            $selectedCategory = $categories[count($categories)-1];
```
TheKeymaster: Pssst.... there is a function called `end()` which may be useful ;).

src/Api/Response/ResponseParser.php
```diff
@@ -201,27 +205,60 @@ protected function parseProducts(SimpleXMLElement $data)
...
         ];
     }
+
+    /**
+     * @param array $requestParams
+     * @return string|null
+     */
+    private function getSelectedCategoryName(array $requestParams)
+    {
+        if (!$selectedCategory = isset($requestParams['attrib']['cat']) ? $requestParams['attrib']['cat'][0] : null) {
```
TheKeymaster: Why are we still checking for falsiness inside `$requestParams['attrib']['cat'][0]`?

TheKeymaster: I think it's cleaner to say `$selectedCat = $requestParams['attrib']['cat'][0] ?? null`.

src/Api/Response/ResponseParser.php
```diff
@@ -201,27 +205,60 @@ protected function parseProducts(SimpleXMLElement $data)
...
+            'selectedCategoryName' => $this->getSelectedCategoryName($requestParams),
+            'selectedVendorName' => $this->getSelectedVendorName($requestParams)
         ];
     }
+
+    /**
+     * @param array $requestParams
+     * @return string|null
+     */
+    private function getSelectedCategoryName(array $requestParams)
```
TheKeymaster: Add return types.

src/Api/Response/ResponseParser.php
```diff
@@ -201,27 +205,60 @@ protected function parseProducts(SimpleXMLElement $data)
...
+        }
+
+        return $selectedCategory;
+    }
+
+    /**
+     * @param array $requestParams
+     * @return string|null
+     */
+    private function getSelectedVendorName(array $requestParams)
```
TheKeymaster: Add return types.

src/Api/Response/Response.php
```diff
@@ -90,4 +101,102 @@ public function getResultsCount()
...
+            return $this->translator->trans(
+                'Findologic::Template.queryInfoMessageDefault',
+                [
+                    'hits' => $this->getResultsCount()
+                ]
+            );
+        }
+    }
+
+    private function getCategoryFilterName(): string
```
TheKeymaster: I do not think we need these methods (`getCategoryFilterName`, `getVendorFilterName`) anymore.

src/Api/Response/Response.php
```diff
@@ -90,4 +101,102 @@ public function getResultsCount()
 
         return $this->data[self::DATA_RESULTS]['count'];
     }
-}
\ No newline at end of file
+
+    public function getQueryInfoMessage(): string
+    {
+        if (!isset($this->data[self::DATA_QUERY_INFO_MESSAGE])) {
```
TheKeymaster: Is this a scenario? If so, when can it happen and how can it be reproduced?

src/Api/Response/Response.php
darius-adeoweb: Filter name can't be retrieved from the request. In the request there is filter codes and values, but not Display name. 
For that reason, I am parsing the response for the Display name.

TheKeymaster: You are totally right.

src/Api/Response/Response.php
darius-adeoweb: Used a wrong function call. Fixed to check for empty array. 
There is a case when parsing DATA_QUERY_INFO_MESSAGE, that it can be an empty array, if no query data present in response.

src/Api/Response/ResponseParser.php
darius-adeoweb: Using `: ?string` throws an error when provisioning plugin on Plentymarkets system. 
`233: UndefError SyntaxError syntax error, unexpected '?'`

Or is there some other way to provide that return type can be null as well, without breaking plugin provisioning?

TheKeymaster: image
.. can't Plentymarkets update their PHP version to something that is actively maintained. PHP 7.0 doesn't even get security fixes anymore:
image

Let's leave out the typehint then.

src/Api/Response/Response.php
```diff
@@ -90,4 +101,104 @@ public function getResultsCount()
...
+            return $this->translator->trans(
+                'Findologic::Template.queryInfoMessageDefault',
+                [
+                    'hits' => $this->getResultsCount()
+                ]
+            );
+        }
+    }
+
+    private function getCategoryFilterName(): string
```
TheKeymaster: To be honest I'd rather have `null` returned than an empty string. You can drop the return type.
Meaning `null` is returned if the category filter name can not be found.

src/Api/Response/ResponseParser.php
```diff
@@ -25,17 +26,20 @@ class ResponseParser
...
         $this->logger = $loggerFactory->getLogger(Plugin::PLUGIN_NAMESPACE, Plugin::PLUGIN_IDENTIFIER);
     }
 
     /**
+     * @param HttpRequest $request
      * @param string $responseData
      * @return Response
      */
-    public function parse($responseData)
+    public function parse($request, $responseData)
```
TheKeymaster: `$request` can be typehinted.

src/Api/Response/ResponseParser.php
darius-adeoweb: Yup, same reaction :/

PR #143, opened 2020-02-03
PLENTY-190 Fix Mobile Smart Suggest clicks

TheKeymaster:


resources/views/content/scripts.twig
```diff
@@ -1,18 +1,20 @@
 <script type="text/javascript">
-    document.addEventListener('flAutocompleteReady', function (e) {
-        var originalSelectItem = e.detail.flAutocomplete.selectItem;
-        e.detail.flAutocomplete.selectItem = function (event, ui) {
-            originalSelectItem.call(e.detail.flAutocomplete, event, ui);
+    window.flSearchSubmitEvent = document.createEvent('Event');
+    window.flSearchSubmitEvent.initEvent('fl-search-submit', true, true);
```
dblaichinger: Use up-to-date method for events since `initEvent` is deprecated

resources/views/content/scripts.twig
```diff
@@ -1,18 +1,20 @@
...
+    document.addEventListener('fl-search-submit', function (e) {
+        e.stopImmediatePropagation();
```
dblaichinger: Check if this is necessary if you set `bubbles = false` on the event itself.

resources/views/content/scripts.twig
TheKeymaster: We would lose support for IE 11 if we'd use `CustomEvent` instead of `initEvent`.

CustomEvent support:
customevent

initEvent() support:
initevent

PR #147, opened 2020-02-21
PLENTY-188 Redirect to product page if one search result 1

darius-adeoweb:


src/Services/SearchService.php
```diff
@@ -249,4 +260,43 @@ private function filterInvalidVariationIds(array $ids)
             return $result['total'] > 0;
         }));
     }
+
+    private function isRedirectToProductPage(array $productsIds): bool
```
TheKeymaster: More like `shouldRedirectToProductDetailPage`.

src/Services/SearchService.php
```diff
@@ -235,6 +241,11 @@ public function aliveTest()
         return $response === Plugin::API_ALIVE_RESPONSE_BODY;
     }
 
+    public function handleProductRedirectUrl(string $url)
```
TheKeymaster: Add a return type. E.g. `void`.

src/Services/SearchService.php
```diff
@@ -249,4 +260,43 @@ private function filterInvalidVariationIds(array $ids)
...
+        $type = !empty($dataQueryInfoMessage['didYouMeanQuery'])
+            ? 'did-you-mean' : $dataQueryInfoMessage['queryStringType'];
+
+        return $type !== 'corrected' && $type !== 'improved';
+    }
+
+    /**
+     * @return string|null
+     */
+    private function getProductRedirectUrl(int $productId)
```
TheKeymaster: Add the return type. E.g. `?string`.

TheKeymaster: This actually gives you the product detail page url, not a redirect URL, because you could theoretically use it for any other purpose than redirecting. Rename this to `getProductDetailUrl`.

src/Services/SearchService.php
```diff
@@ -275,7 +275,10 @@ private function shouldRedirectToProductDetailPage(array $productsIds): bool
         return $type !== 'corrected' && $type !== 'improved';
     }
 
-    private function getProductDetailUrl(int $productId): ?string
+    /**
+     * @return string|null
```
TheKeymaster: Why did adding `?string` not work?

darius-adeoweb: Using : ?string throws an error when provisioning plugin on Plentymarkets system.
233: UndefError SyntaxError syntax error, unexpected '?'

Same problem as in the https://github.com/findologic/plugin-plentymarkets-ceres-api/pull/142

TheKeymaster: Ah yes I have totally forgotten that. Damn.

PR #148, opened 2020-02-21
PLENTY-158 Add gulp git hook

darius-adeoweb:


.pre-commit
```diff
@@ -0,0 +1,25 @@
...
+        exit 0;
+    fi
+
+    # Changes not compiled
+    gulp js
+    gulp sass
+
+    if git diff --name-only --diff-filter=ACM | grep "resources/js/dist\|resources/css"
+    then
+        printf "ERROR: JS/CSS was not compiled before pushing. Run gulp\n"
```
TheKeymaster: I'd like to have `gulp` in quotation marks. This makes it more obvious that this is a command.

TheKeymaster: The moment you've commited, gulp was already executed. You could also just mention that these files need to be added to the current commit in order to be commited.

gulpfile.js
```diff
@@ -68,4 +69,9 @@ gulp.task('sass', function () {
         .pipe(gulp.dest('resources/css'));
 });
 
+gulp.task('install-hooks', function () {
```
TheKeymaster: Add that this command needs to be executed before developing to the development section in the `README.md`.

darius-adeoweb: I have added the instructions to the installing section in the README.md now. If development is better, I will update the pull request.

PR #155, opened 2020-03-03
PLENTY-200 On navigation show plentymarkets sorting

darius-adeoweb:


src/Services/Search/ParametersHandler.php
```diff
@@ -65,24 +65,44 @@ public function handlePaginationAndSorting($search, $request)
         $config = $this->getConfig();
 
         $isSearch = strpos($request->getUri(), '/search') !== false;
+        $isFiltersSet = strpos($request->getUri(), '?attrib') !== false;
```
TheKeymaster: Use `$request->all()` and check if there is an `attrib` key as parameter.

src/Services/Search/ParametersHandler.php
```diff
@@ -65,24 +65,44 @@ public function handlePaginationAndSorting($search, $request)
...
+
+        return $this->getFindologicSortingOptions($isSearch);
+    }
+
+    public function getPlentymarketsSortingOptions(): array
+    {
+        /** @var CeresConfig $config */
+        $config = $this->getConfig();
+
+        $returnArray = [];
```
TheKeymaster: `$returnArray` is not self explanatory, try to find a better name that suites the value of the variable.

tests/Services/Search/ParametersHandlerTest.php
```diff
@@ -19,6 +19,53 @@ public function getSortingOptionsProvider()
         return [
             'Sorting options for category page' => [
                 false,
+                false,
```
TheKeymaster: When we are at it, please add keys to these data providers, so you can read the dataProvider without the need of the associated test. In this case:

```php
return [
    'Sorting options for category page' => [
        'isSearch' => false,
        'hasFilters' => false,
        // . . . 
    ]
]
```

TheKeymaster: Please do this for all touched data providers.

tests/Services/Search/ParametersHandlerTest.php
```diff
@@ -280,10 +383,18 @@ public function testHandlePaginationAndSorting(
...
 
+        $returnedUri = 'test.com';
+
         if ($isSearch) {
-            $requestMock->expects($this->once())->method('getUri')->willReturn('test.com/search');
+            $returnedUri .= '/search';
         }
 
+        if ($isFiltersSet) {
+            $returnedUri .= '?attrib';
```
TheKeymaster: To make this more realistic, add some fake value to it. Only having the key is not really realistic.

PR #162, opened 2020-03-10
PLENTY-213 Fix ceres search bar permanent

TheKeymaster:


TheKeymaster:
Ceres 5 fixed this automatically. No need to investigate this further.

PR #171, opened 2020-03-16
PLENTY-219_configure_price_as_text

TheKeymaster:


resources/views/ItemList/Components/Filter/ItemFilterPrice.twig
```diff
@@ -0,0 +1,32 @@
+<script type="x/template" id="vue-findologic-item-price">
```
dblaichinger: Just for my understanding, why didn't we need this for Ceres 4.6? Did Plenty remove this template in 5.0?

dblaichinger: And was this copied or did you write it yourself?

tests/Mocks/PlentyRequestMock.php
```diff
@@ -198,4 +198,11 @@ public function getQueryString(): string
     public function getMethod(): string
     {
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function getLocale(): string
+    {
+    }
```
dblaichinger: Is this related to the price filter? Or a change which sneaked in?

.travis.yml
```diff
@@ -13,4 +13,4 @@ install:
   - php composer.phar install
 
 script:
-  - php vendor/bin/phpunit -c tests/phpunit.xml
\ No newline at end of file
+  - composer test
```
dblaichinger: Keeping the travis yml clean and simple, I like it :)

resources/views/ItemList/Components/Filter/ItemFilterPrice.twig
TheKeymaster: > Just for my understanding, why didn't we need this for Ceres 4.6? Did Plenty remove this template in 5.0?

Plenty now uses `.vue` files. We manually need to override these to be able to use them as before.

> And was this copied or did you write it yourself?

The code was copied straight from plenty. It will be removed once we also switch to `.vue` files.

tests/Mocks/PlentyRequestMock.php
TheKeymaster: I have updated the dependency `plentymarkets/plugin-interface`. Since the abstract class (which this one extends from) added this method additionally as `abstract` I also needed to add it here.

.travis.yml
TheKeymaster: :+1:

resources/views/ItemList/Components/Filter/ItemFilterPrice.twig
dblaichinger: Please create a story with all necessary information to convert our twig files to vue. Also mentioning the files which can be removed like this one. When this is done this discussion can be resolved

PR #172, opened 2020-03-16
PLENTY-222_dont_show_two_checkboxes_vendor

TheKeymaster:


resources/scss/views/ItemList/Components/Filter/item-filter-list.scss
```diff
@@ -53,6 +53,10 @@
                                 opacity: 1;
                                 padding-left: 0;
                             }
+
+                            span::before {
+                                display: none;
+                            }
```
dblaichinger: Fix in CSS is fine when there is no other way but do you know the reason why Plenty even renders two checkboxes?

TheKeymaster: Short answer: The selector sucks.
Long answer: The label element has two elements; the image and the label. Vue renders them both. The `::before` selector is even shown if the element itself is hidden via `v-show`.

Conclusion: I have now added a class to hide the checkbox for certain elements, as we can not access `::before` via html.

dblaichinger: > The ::before selector is even shown if the element itself is hidden via v-show.

Same with `v-if`?

TheKeymaster: I'd need to check.. if it works out I will close this, otherwise I'ill add a comment.

TheKeymaster: So `v-if` does also not work and manually only applying the styles to the affected element does also not work. I already spent much time for that (as can be seen by the amount of commits) for such a small thing. I'd suggest leaving it like that.

PR #173, opened 2020-03-16
PLENTY-218_fix_dropdown_border

TheKeymaster:


resources/scss/views/ItemList/Components/Filter/item-dropdown.scss
```diff
@@ -24,9 +24,8 @@
             width: 100%;
             z-index: 2;
             background-color: $white;
-            border-bottom: 1px solid;
-            border-left: 1px solid;
-            border-right: 1px solid;
+            border: 1px solid #000;
+            border-top: transparent;
```
dblaichinger: I think the initial value is `none`

TheKeymaster: Setting it to `none` would mean that we do not override the `1px solid #000`, which is not what we want. I also thought about that but it doesn't work.

dblaichinger: It doesn't override it? Wow I was not expecting this behaviour. Alright then leave it as it is.

PR #174, opened 2020-03-17
PLENTY-220_fix_shopbuilder_filter_button

TheKeymaster:


resources/scss/shop-builder.scss
```diff
@@ -0,0 +1,6 @@
+// This file overwrites standard plenty shop builder styles. All styles found in this file only
+// have an effect when using the shop builder. The styles are always loaded, even if the shop builder is not used.
```
dblaichinger: > All styles found in this file only have an effect when using the shop builder. The styles are always loaded, even if the shop builder is not used.

That is a bit misleading, if the styles are always loaded even without shop builder, how can those style only have effect on the shop builder? Are those classes only available when using shop builder?

TheKeymaster: > Are those classes only available when using shop builder?

Congratulations, you are a millionaire! Yes those classes are only available when using Shop Builder and I want to keep it that way for future issues. Everything that is only-shopbuilder should stay in here and should only have an effect when using only-shopbuilder.

dblaichinger: Please update your comment to make it more clear, when you stumble across this file the first time its confusing.

TheKeymaster: I hope it's more clear now.

PR #180, opened 2020-03-26
PLENTY-230 Fix sdym with query info message

TheKeymaster:


src/Api/Response/Response.php
```diff
@@ -172,6 +161,52 @@ public function getQueryInfoMessage(): string
...
+        }
+
+        $type = !empty($dataQueryInfoMessage['didYouMeanQuery'])
+            ? 'did-you-mean' : $dataQueryInfoMessage['queryStringType'];
+        $alternativeQuery = !empty($dataQueryInfoMessage['didYouMeanQuery'])
+            ? $dataQueryInfoMessage['didYouMeanQuery']
+            : $dataQueryInfoMessage['currentQuery'];
+        $originalQuery = !empty($dataQueryInfoMessage['didYouMeanQuery'])
+            ? $dataQueryInfoMessage['currentQuery']
+            : $dataQueryInfoMessage['originalQuery'];
```
dblaichinger: This is hard to read and three checks for `!empty($dataQueryInfoMessage['didYouMeanQuery'])`.
Just check it once and set all default values when empty. Maybe a comment would help or renaming variables since the logic is not clear when just reading it.

src/Api/Response/Response.php
```diff
@@ -172,6 +161,52 @@ public function getQueryInfoMessage(): string
...
+
+        if ($type === 'corrected') {
```
dblaichinger: I would prefer using `switch($type)` here

tests/Api/Response/ResponseTest.php
```diff
@@ -122,10 +122,10 @@ public function getQueryInfoMessageProvider(): array
                     'selectedVendorName' => null
                 ],
                 [],
-                'Findologic::Template.correctedQuery',
+                'Findologic::Template.queryInfoMessageQuery',
                 [
-                    'alternativeQuery' => 'Test',
-                    'originalQuery' => 'OriginalTest'
+                    'query' => 'Test',
+                    'hits' => 0
```
dblaichinger: Is there no new test for the changed behavior?
1. The SmartDidYouMeanCotainer now calls `->getSmartDidYouMean()`
1. Two lines of output instead of one.
1. One sdym case was broken, as stated in the story: "Fix the did you mean case."

src/Api/Response/Response.php
TheKeymaster: Good improvement, nice idea. Thanks!

src/Api/Response/Response.php
```diff
@@ -172,6 +163,54 @@ public function getQueryInfoMessage(): string
...
+    public function getSmartDidYouMean(): string
+    {
+        $dataQueryInfoMessage = $this->getData(self::DATA_QUERY_INFO_MESSAGE);
+
+        if (empty($dataQueryInfoMessage)) {
+            return '';
+        }
+
+        $type = $dataQueryInfoMessage['queryStringType'];
+        $alternativeQuery = $dataQueryInfoMessage['currentQuery'];
```
dblaichinger: Why isn't the key called `alternativeQuery`? Isn't this only the alternative query when SDYM is triggered and during normal search the query?

src/Api/Response/Response.php
```diff
@@ -110,29 +110,20 @@ public function getQueryInfoMessage(): string
...
-                ]
-            );
-        } else if ($type === 'improved') {
+        $type = $dataQueryInfoMessage['queryStringType'];
+        $alternativeQuery = $dataQueryInfoMessage['currentQuery'];
+
+        if (!empty($dataQueryInfoMessage['didYouMeanQuery'])) {
+            $type = 'did-you-mean';
+            $alternativeQuery = $dataQueryInfoMessage['didYouMeanQuery'];
+        }
```
dblaichinger: Good improvement, way more readable now :)

src/Api/Response/Response.php
```diff
@@ -172,6 +163,54 @@ public function getQueryInfoMessage(): string
...
+            return '';
+        }
+
+        $type = $dataQueryInfoMessage['queryStringType'];
+        $alternativeQuery = $dataQueryInfoMessage['currentQuery'];
+        $originalQuery = $dataQueryInfoMessage['originalQuery'];
+
+        if (!empty($dataQueryInfoMessage['didYouMeanQuery'])) {
+            $type = 'did-you-mean';
+            $alternativeQuery = $dataQueryInfoMessage['didYouMeanQuery'];
```
dblaichinger: Isn't this the same as in line 116? Can this be combined somehow? If not, create a constant for the string 'did-you-mean' and use it on both places, otherwise we may forgot to update it in one place in the future.

tests/Api/Response/ResponseTest.php
```diff
@@ -253,15 +253,98 @@ public function getQueryInfoMessageProvider(): array
...
+                    'originalQuery' => 'blub',
+                    'didYouMeanQuery' => null,
+                    'currentQuery' => 'blubbergurken',
+                    'queryStringType' => 'corrected',
+                ],
+                'expectedTranslationKey' => 'Findologic::Template.correctedQuery',
+                'expectedTranslationParams' => [
+                    'originalQuery' => 'blub',
+                    'alternativeQuery' => 'blubbergurken'
+                ],
```
dblaichinger: It would be easier to differentiate between different SDYM cases when "real" terms are used for the queries and not blub and blubbergurken. Think how you can see the SDYM cases in the shop e.g. searching for sesel

src/Api/Response/Response.php
TheKeymaster: For searches where the term is improved or corrected, the current query is the one that is found in the response. This would be the alternative query. For the other case it would be did-you-mean.

The key is not called `alternativeQuery`, because we only know that this is either the query that the user sent, or a query that is improved by SDYM.

tests/Api/Response/ResponseTest.php
TheKeymaster: You can not replicate real cases as everything except the corrected query, is based on the query history, which may lead to such a result.
In this case the given search term is always `blub` and everything that is corrected will be `blubbergurken`. I get that we want to be as real in tests as in the real world, but this could be such a case as well.

Nevertheless I think it is a good idea, so I changed it. Sent query is always `sesel` and corrected/improved/did-you-mean is always `sessel`.

src/Api/Response/Response.php
dblaichinger: Ok

tests/Api/Response/ResponseTest.php
dblaichinger: This was just an example is there an english equivalent?

TheKeymaster: Sure. Resolved.

TheKeymaster:
Closed in favor of #178, which has these changes.

PR #186, opened 2020-04-08
PLENTY-227 submit data from the correct search field when there are m…

RimantasAdeoWeb:
…ore than one


resources/views/content/scripts.twig
```diff
@@ -1,20 +1,17 @@
 <script type="text/javascript">
-    window.flSearchSubmitEvent = document.createEvent('Event');
-    window.flSearchSubmitEvent.initEvent('fl-search-submit', false, true);
-    document.addEventListener('fl-search-submit', function (e) {
-        e.preventDefault();
+    function flSearchSubmitEventHandler(event, element) {
```
TheKeymaster, line 2: Rename `element` to `form` or `formElement` or even `searchFormElement` (you can decide), so it is clear that we are dealing with a form.

PR #191, opened 2020-04-28
PLENTY-241_rename_FINDOLOGIC_to_Findologic

dblaichinger:


config.json
```diff
@@ -1,8 +1,8 @@
 {
   "formatVersion": 1,
   "menu": {
-    "Config.FINDOLOGICSearchTab": {
-      "label": "Config.FINDOLOGICSearchTab",
+    "Config.FindologicSearchTab": {
```
TheKeymaster, line 4: Have you checked if Plentymarkets manages to save these values? Otherwise upgrading would result in all configuration to be removed.

meta/documents/changelog_en.md
```diff
@@ -37,7 +37,7 @@
 
 ### Added
 
-* [PLENTY-192] Plentymarkets tags are now supported. FINDOLOGIC will automatically filter by the given tag on tag pages.
+* [PLENTY-192] Plentymarkets tags are now supported. Findologic will automatically filter by the given tag on tag pages.
```
TheKeymaster, line 40: If you change the changelogs here, they should also be changed in the [Releases on GitHub](https://github.com/findologic/plugin-plentymarkets-ceres-api/releases).

config.json
dblaichinger: Yes click tested and all values are set as before, all container connections are there, nothing changed.
Screen Shot 2020-04-28 at 17 58 31
Screen Shot 2020-04-28 at 17 59 15

meta/documents/changelog_en.md
dblaichinger: Added post-deploy task

composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "ac26209d38878a5f59c2b8c104acff69",
```
TheKeymaster, line 7: Please revert the commit https://github.com/findologic/plugin-plentymarkets-ceres-api/pull/191/commits/0bd17604b824ce7217d0539f5f6e4e1dc8b73ce3, as in this commit, all packages were updated, which was probably not your plan :D.

PR #200, opened 2020-08-06
PLENTY-225_send_filters_only_once

TheKeymaster:


src/Api/Request/Request.php
```diff
@@ -143,7 +143,13 @@ public function setAttributeParam($key, $value)
             $this->params[Plugin::API_PARAMETER_ATTRIBUTES] = [];
         }
 
-        $this->params[Plugin::API_PARAMETER_ATTRIBUTES][$key] = $value;
+        // Remove duplicate entries, since the plenty request may give them as duplicates. Additionally
+        // check that min/max is not set, because the range-slider may have the same min and max value.
+        if (is_array($value) && (!isset($value['min']) && !isset($value['max']))) {
```
dblaichinger, line 148: With this logic we don't remove duplicates when the price slider is used, but what happens when several filters are in use and one of them is the price slider?
I guess a solution to exclude any min and max values from the `array_unique` would be more specific.

tests/Api/Request/ParametersBuilderTest.php
```diff
@@ -230,7 +230,32 @@ public function setSearchParamsProvider()
...
+                    ],
+                    'attrib' => [
+                        'vendor' => [
+                            'A & C Design',
+                        ]
+                    ],
+                    'order' => 'price ASC',
+                    'count' => 10
+                ]
+            ],
```
dblaichinger, line 258: Can you please also test any logic related to min and max.

src/Api/Request/Request.php
```diff
@@ -143,7 +143,13 @@ public function setAttributeParam($key, $value)
             $this->params[Plugin::API_PARAMETER_ATTRIBUTES] = [];
         }
 
-        $this->params[Plugin::API_PARAMETER_ATTRIBUTES][$key] = $value;
+        // Remove duplicate entries, since the plenty request may give them as duplicates. Additionally
+        // check that min/max is not set, because the range-slider may have the same min and max value.
```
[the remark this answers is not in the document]
TheKeymaster: Honestly this is just so slow and they will, as I know plenty, never fix that.

src/Api/Request/Request.php
TheKeymaster: This is already per-filter. Basically it is like first vendor then price, etc. So we are only getting the unique values per-filter already. Added another test to reflect that.

tests/Api/Request/ParametersBuilderTest.php
TheKeymaster: There is already a test that tests for same min and max value. I now added a test that checks both min and max is persisted but same filter values are removed.

PR #202, opened 2020-08-06
PLENTY-254 redirect with encoded search params instead of submiting form

RimantasAdeoWeb:


resources/views/content/scripts.twig
```diff
@@ -7,10 +7,9 @@
             action = '/' + App.language + '/search';
         }
 
-        formElement.setAttribute('action', action);
-        formElement.submit();
+        query = document.querySelector("input.search-input").value;
```
TheKeymaster, line 10: This variable was not declared before, but since this is JavaScript, it allows it anyway. Please declare it using `var`.

resources/views/content/scripts.twig
```diff
@@ -7,10 +7,9 @@
...
 
-        return true;
+        document.location = action + '?query=' + query
```
TheKeymaster, line 12: Missing a `;` at the end.

PR #204, opened 2020-08-11
PLENTY-260_fix_search_result_with_io

TheKeymaster:


tests/Services/SearchServiceTest.php
```diff
@@ -155,6 +156,177 @@ public function testRedirectToProductPageOnDoSearch(
...
+                    'total' => 1,
+                    'documents' => [
+                        [
+                            'id' => 1011
+                        ]
+                    ]
+                ],
+                'attributes' => []
+            ],
+            'One product found and query string type is improved' => [
```
dblaichinger, line 210: Isn't that the same scenario as ` 'One product found and query string type is corrected'`?

dblaichinger: It seems that query string type differentiation is not making any difference, since `'One product found'` only differs from that number of variation ids

tests/Services/SearchServiceTest.php
```diff
@@ -155,6 +156,177 @@ public function testRedirectToProductPageOnDoSearch(
...
+        $requestMock->expects($this->any())
+            ->method('all')
+            ->willReturn($attributes);
+
+        /** @var Request|HttpRequest|MockObject $searchRequestMock */
+        $searchRequestMock = $this->getMockBuilder(Request::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        // Ensure that the given category is null. This way we are sure that the category is being ignored.
```
dblaichinger, line 322: But what about scenario with `'cat' => 'Blubbergurken'`? Can you test this category isn't ignored?

tests/Services/SearchServiceTest.php
TheKeymaster: Yeah they were actually the same. I actually copied that provider and removed the stuff that I didn't need. This caused that duplicated scenarios. They're removed now ^^.

PR #205, opened 2020-09-04
PLENTY-262_fix_redirect_for_page

TheKeymaster:


tests/Services/SearchServiceTest.php
```diff
@@ -219,6 +219,47 @@ public function redirectToProductPageOnDoSearchProvider(): array
                 'redirectUrl' => '/test-product/a-11',
                 'attributes' => []
             ],
+            'One product found on second page' => [
```
dblaichinger, line 222: It's not clear as this case is in `redirectToProductPageOnDoSearchProvider`. Maybe call it 'One product found on second page should not redirect to product detail' or similar.

tests/Services/SearchServiceTest.php
```diff
@@ -219,6 +219,47 @@ public function redirectToProductPageOnDoSearchProvider(): array
...
+                        ]
+                    ]
+                ],
+                'shopUrl' => 'https://www.test.com',
+                'dataQueryInfoMessage' => [
+                    'queryStringType' => 'notImprovedOrCorrected'
+                ],
+                'redirectUrl' => null,
+                'attributes' => [
+                    'page' => 2
```
dblaichinger, line 260: Maybe add a test case with page = 1 as you explicitly check for it.

dblaichinger (changes requested):

PR #210, opened 2020-10-05
PLENTY-274 fix_smart_suggest_clicks

TheKeymaster:


resources/views/content/scripts.twig
```diff
@@ -7,7 +16,11 @@
             action = '/' + App.language + '/search';
         }
 
-        document.location = action + '?query=' + formElement.querySelector('input.search-input').value;
+        if (formElement.querySelector('input.search-input').value.trim() !== '') {
+            handleSearchQuery(formElement, action);
+        } else {
+            handleSmartSuggestClick(formElement, action);
```
dblaichinger, line 22: Do we also need to handle Guided Shoppping with empty query?

dblaichinger (commented):
Code lgtm, one question.

resources/views/content/scripts.twig
TheKeymaster: The Shopping Guide works just fine.

dblaichinger (approved):

PR #212, opened 2020-10-28
PLENTY-277 handle shop keys with no language code

RimantasAdeoWeb:


tests/Components/PluginConfigTest.php
```diff
@@ -80,40 +80,66 @@ public function testPushIsForwardedToConfigRepository()
...
+                'de',
+                'ABC123'
+            ],
+            'It returns null when no generic shopkey and no shopkey for current language are configured 1' => [
+                "de:ABC123\njp:DEF456",
+                'en',
+                null
+            ],
+            'It returns null when no generic shopkey and no shopkey for current language are configured 2' => [
+                "en:ABC123",
```
TheKeymaster, line 109: You do not need double quotes for this string.

tests/Components/PluginConfigTest.php
```diff
@@ -80,40 +80,66 @@ public function testPushIsForwardedToConfigRepository()
...
 
-    public function testGetShopKeyReturnsNullIfShopKeyIsEmpty()
+    /**
+     * @dataProvider getShopKeyProvider
+     *
+     * @param string $configRepositoryGetReturnValue
+     * @param string $currentLang
+     * @param string|null $returnValue
+     */
+    public function testGetShopKey($configRepositoryGetReturnValue, $currentLang, $returnValue)
```
TheKeymaster, line 133: `$returnValue` and `$configRepositoryGetReturnValue` are odd names, as a "return value" could be literally anything.

* `$configRepositoryGetReturnValue` => `$configuredShopkeys` (maybe even `$raw...`, to make it clear this value has not been touched yet)
* `$returnValue` => `$expectedShopkey`

Naming the variables what they contain, instead of where they will be used, hides implementation specific details, which makes this just way easier to read and to understand.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #213, opened 2020-10-30
PLENTY-257 when search has one result, redirect to the variation whic…

RimantasAdeoWeb:
…h has an identifier matching the search query

### DIFF
--- a/src/Services/SearchService.php
+++ b/src/Services/SearchService.php
@@ -103,7 +103,7 @@ class SearchService implements SearchServiceInterface
 
     public function getSearchFactory(): VariationSearchFactory
     {
-        return pluginApp( VariationSearchFactory::class);
+        return pluginApp(VariationSearchFactory::class);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Services/SearchService.php
```diff
@@ -101,13 +101,9 @@ public function getItemSearchService()
...
 
-    public function getSearchFactory(array $ids): VariationSearchFactory
+    public function getSearchFactory(): VariationSearchFactory
     {
-        $searchFactory = pluginApp( VariationSearchFactory::class);
-
-        $searchFactory->hasVariationIds($ids);
-
-        return $searchFactory;
+        return pluginApp( VariationSearchFactory::class);
```
TheKeymaster, line 106: That one space can be removed.

src/Services/SearchService.php
```diff
@@ -138,16 +134,17 @@ public function getResults()
     public function doSearch(HttpRequest $request, ExternalSearch $externalSearch)
     {
         $results = $this->search($request, $externalSearch);
-        $productsIds = $this->filterInvalidVariationIds($results->getVariationIds());
 
-        if ($this->shouldRedirectToProductDetailPage($productsIds, $request)) {
-            if ($redirectUrl = $this->getProductDetailUrl($productsIds[0])) {
+        $variationIds = $this->filterInvalidVariationIds($results->getVariationIds());
```
TheKeymaster, line 138: Good idea to rename this parameter :+1:

src/Services/SearchService.php
```diff
@@ -306,24 +303,57 @@ private function shouldRedirectToProductDetailPage(array $productsIds, HttpReque
...
 
+        $productData = $result['documents'][0]['data'];
         $urlPath = $productData['texts'][0]['urlPath'];
         $itemId = $productData['item']['id'];
+        $variationId = $this->getVariationIdForRedirect($query, $result['documents']);
+
+        return sprintf('/%s_%s_%s', $urlPath, $itemId, $variationId);
+    }
+
+    private function getVariationIdForRedirect(string $query, array $documents)
```
TheKeymaster, line 330: 1. The return type is not clear. Add one or add it in the docblock.
1. The method name does not suggest, that the main variation id will be returned, in case no appropriate variation can be found. Either add a docblock or update the method name.

src/Services/SearchService.php
```diff
@@ -306,24 +303,57 @@ private function shouldRedirectToProductDetailPage(array $productsIds, HttpReque
...
+    {
+        $query = strtolower($query);
```
TheKeymaster, line 332: Try to not override existing variables in the same method. This would be a `$lowercasedQuery`, or you could call the given argument `$rawQuery`, but they should not override eachother.

TheKeymaster (changes requested):
Only a few comments in the actual implementation, tests look good!

TheKeymaster (approved):
Looks good!

PR #215, opened 2020-11-11
PLENTY-159 filter invalid products only for Ceres versions below 5.0.3

RimantasAdeoWeb:


src/Services/SearchService.php
```diff
@@ -247,6 +259,34 @@ public function handleProductRedirectUrl(string $url)
         header('Location: ' . $url);
     }
 
+    /**
+     * Since Ceres version 5.0.3 invalid ids are filtered by default.
+     * Custom version checking as Plentymarkets forbids using the version_compare function
```
TheKeymaster, line 264: The sentence `Custom version checking as Plentymarkets forbids using the version_compare function`  is implementation specific, therefore please put it as inline doc before declaring the `$versionParts` variable.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #216, opened 2020-11-12
PH-16_fix_language_redirect (PLENTY-276)

TheKeymaster:


src/Services/SearchService.php
```diff
@@ -364,4 +367,31 @@ private function getVariationIdForRedirect(string $query, array $documents)
 
         return $mainVariationId;
     }
+
+    /**
+     * @see \IO\Extensions\Filters\URLFilter::buildItemURL (source)
+     */
+    private function buildItemURL($itemData, $withVariationId = true): string
```
dblaichinger, line 374: Passing the variation id directly would make the flag unnecessary, right?

src/Services/SearchService.php
```diff
@@ -364,4 +367,31 @@ private function getVariationIdForRedirect(string $query, array $documents)
...
+
+        $includeLanguage = Utils::getLang() !== Utils::getDefaultLang();
+        if ($variationId === null || $variationId <= 0) {
+            return $urlBuilderRepository->buildItemUrl($itemId)->toRelativeUrl($includeLanguage);
+        } else {
+            $url = $urlBuilderRepository->buildVariationUrl($itemId, $variationId);
+
+            return $url->append(
+                $urlBuilderRepository->getSuffix($itemId, $variationId, $withVariationId)
+            )->toRelativeUrl($includeLanguage);
```
dblaichinger, line 394: Really strange to use a boolean here and not a language string, but it's how plenty defines it.

tests/Services/SearchServiceTest.php
```diff
@@ -993,4 +1048,70 @@ public function redirectToProductPageOnDoSearchProvider(): array
...
+                return $urlQueryMock;
+            });
+
+        $urlQueryMock->expects($this->any())->method('toRelativeUrl')
+            ->willReturnCallback(function ($includeLanguage) use ($urlQueryMock, $defaultLanguage, $language) {
+                if (!$includeLanguage || $language === $defaultLanguage) {
+                    return parse_url($urlQueryMock->url, PHP_URL_PATH);
+                }
+
+                return '/' . $language . parse_url($urlQueryMock->url, PHP_URL_PATH);
```
dblaichinger, line 1101: The parse_url result could be extracted to a variable.

tests/autoload.php
```diff
@@ -10,4 +10,22 @@
...
 $classLoader->add('Findologic\Tests', __DIR__);
 
-$classLoader->register();
\ No newline at end of file
+/**
+ * Add class instances that may be defined in tests. In production this is defined by Plentymarkets.
+ *
+ * @var array<string, object> $classInstances
+ */
+global $classInstances;
```
dblaichinger, line 18: Necessary to define `global $classInstances;` twice?

dblaichinger (changes requested):

src/Services/SearchService.php
TheKeymaster: I want to stick with the plenty implementation, so it can be updated easily in the future if it changes.

src/Services/SearchService.php
TheKeymaster: It is what it is.

tests/autoload.php
TheKeymaster: Yes since they're in different scopes. Here I am defining it as global and initializing it as an array. In the method `pluginApp` I just say that I want to get the global variable. If I wouldn't define it, it may be null, which I want to avoid.

dblaichinger (approved):

PR #218, opened 2020-11-26
PLENTY-266 page builder filter widget compatibility

RimantasAdeoWeb:


src/Api/Response/Parser/FiltersParser.php
```diff
@@ -70,7 +73,28 @@ public function parseForWidgets(SimpleXMLElement $data): array
...
+                case 'price':
+                    $filter['type'] = 'price';
+                    break;
+                default:
+                    $filter['type'] = 'dynamic';
+                    break;
+            }
+        }
+
+        unset($filter);
```
TheKeymaster, line 97: PHP should automatically garbage collect this, no need of manually unsetting that.

src/Api/Response/Parser/FiltersParser.php
```diff
@@ -39,9 +39,12 @@ public function __construct(LibraryCallContract $libraryCallContract, Image $col
...
     }
 
-    public function parse(SimpleXMLElement $data): array
+    /**
+     * @param SimpleXMLElement|null $data
+     */
+    public function parse($data): array
     {
-        if (!$data) {
+        if (!$data instanceof SimpleXMLElement) {
```
TheKeymaster, line 47: Shouldn't this be typehinted like `parseForWidgets` instead? They receive the same data anyway.

TheKeymaster: If one isn't of type `SimpleXMLElement` the other wouldn't be of type `SimpleXMLElement` either. Either you handle both like that, or you add a typehint to both. Not sure if there may be a case where `$data` wouldn't be of `SimpleXMLElement`, but since you added this check, you probably have more insight into this.

tests/Api/Response/Parser/FiltersParserTest.php
```diff
@@ -1119,4 +1135,160 @@ public function parseDataProvider()
...
+    {
+        return [
+            'typesGetMappedCorrectlyIfValuesExist' => [
+                [
+                    [
+                        'id' => 'vendor',
+                        'name' => 'Manufacturer',
+                        'type' => 'image',
+                        'values' => [
+                            ['id' => 1]
```
TheKeymaster, line 1161: Similar to `parseDataProvider`, shouldn't this contain more as just the id? Please try to use more realistic data.

TheKeymaster (changes requested):

src/Api/Response/Parser/FiltersParser.php
RimantasAdeoWeb: Not for garbage collection, just a good practice when working with foreach loops and references. https://www.php.net/manual/en/control-structures.foreach.php "Warning Reference of a $value and the last array element remain even after the foreach loop. It is recommended to destroy it by unset(). "

TheKeymaster: Ok, since I really do not like unsetting such things manually, it may make sense to move the whole foreach into a separate method and create an array, without using references. Build the array yourself and return it then. I thought of something like that:

```php
function callThisSomethingReasonable(array $filters)
{
    $result = [];
    foreach ($filters as $filter) {
        // switch { ...
        // case: ...
            // $filter['manipulated'] = 'manipulated';
            $result[] = $filter;
    }
}
```

This is cleaner and prevents such things from even happening in the first place.

tests/Api/Response/Parser/FiltersParserTest.php
RimantasAdeoWeb: It does contain more than just the ID but it really isn't relevant for the test and the full objects are quite large. Do I just copy everything over?

TheKeymaster: Each filter containing 1-2 values should be sufficient.

src/Api/Response/Parser/FiltersParser.php
RimantasAdeoWeb: I left it in the same function. A separate one seemed unnecessary as this is almost all that parseForWidgets does anyway.

src/Api/Response/Parser/FiltersParser.php
RimantasAdeoWeb: Not sure about such cases but there was a test for that, which was actually failing because this method couldn't handle its input. Rather than adjusting the test, I adjusted the code, seemed safer. Matched the typehints of both methods.

src/Api/Response/Parser/FiltersParser.php
TheKeymaster: That's also fine for me.

TheKeymaster (approved):

PR #219, opened 2020-11-27
Update README

TheKeymaster:


README.md
```diff
@@ -13,41 +13,39 @@
 1. [Deployment & Release](#deployment--release)
 1. [Versioning](#versioning)
 
-## Install
+## Installation
```
TheKeymaster, line 16: You can not have two sections with the same name (the links would not work), therefore I went with `Install`.

README.md
```diff
@@ -73,27 +71,27 @@ Before pushing you may run build to ensure your JS/CSS is built.
...
-Alternatively all tests can be run, by executing
+Alternatively run all tests with
 
 ```
 composer test
 ```
 
 ## Deployment & Release
-1. Update the German and English changelogs in folder `meta/documents`.
+1. Update the german and english change logs in folder `meta/documents`.
```
TheKeymaster, line 87: Countries, nationalities, and languages may always be capitalized.

README.md
dblaichinger: How about `Setup`?

README.md
dblaichinger: Correct please change back.

PR #229, opened 2021-01-19
PLENTY-245 show selected filter name together with the value

RimantasAdeoWeb:


resources/js/src/app/components/itemList/filter/ItemFilterTagList.js
```diff
@@ -41,6 +44,16 @@ Vue.component("item-filter-tag-list", {
 
         resetAllTags() {
             this.removeAllAttribsAndRefresh();
+        },
+
+        getFacetIdNameMap() {
+            let map = {};
```
TheKeymaster, line 50: This may be a `const` instead of `let`.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #231, opened 2021-02-02
PLENTY-299 fix console errors when using page builder filter widgets,…

RimantasAdeoWeb:
… fix conflicts between price filters when a duplicate widget is in the toolbar


TheKeymaster (approved):
Neat Fix!

PR #235, opened 2021-02-12
PLENTY-305 Improve usability of price slider filter

RimantasAdeoWeb:


resources/js/src/app/components/itemList/filter/ItemRangeSlider.js
```diff
@@ -71,6 +87,31 @@ Vue.component("item-range-slider", {
...
+        },
+
+        setCustomValidationMessage() {
+            this.$el.querySelectorAll('input.fl-range-input[data-custom-validation-message]').forEach((input) => {
+                // Must be reset before the validity check as existence of custom validity counts as a validation error.
+                input.setCustomValidity('');
+
+                if (!input.checkValidity()) {
+                    input.setCustomValidity(input.dataset.customValidationMessage);
+                }
```
TheKeymaster, line 113: Overall nice usage of the available APIs!

resources/js/src/app/mixins/url.js
```diff
@@ -418,6 +422,16 @@ export default {
             delete params[Constants.PARAMETER_PAGE];
             delete params[Constants.PARAMETER_ATTRIBUTES];
             document.location.search = '?' + $.param(params);
+        },
+
+        getFacetIdInfoMap() {
```
TheKeymaster, line 427: `facetId` looks like something internal of the method to me. I do not think it is crucial for the user of this method, to know how the info map is generated.

resources/js/src/app/mixins/url.js
```diff
@@ -238,9 +238,13 @@ export default {
                 }
 
                 if (filter === 'price' || this.isRangeSliderFilter(attributes[filter])) {
+                    let facetInfo = this.getFacetIdInfoMap();
+
+                    let unit = (facetInfo[filter] && facetInfo[filter].unit) ? ' ' + facetInfo[filter].unit : '';
```
TheKeymaster, line 243: The unit itself shouldn't contain the space. This should be done, when assigning the name

TheKeymaster (changes requested):

resources/js/src/app/components/itemList/filter/ItemRangeSlider.js
```diff
@@ -71,6 +87,31 @@ Vue.component("item-range-slider", {
...
             }
+        },
+
+        fixDecimalSeparator(value) {
+            if (typeof value === 'number') {
+                value = value.toString();
+            }
+
+            if (value.indexOf('.') > -1) {
+                value = value.replace(',', '');
```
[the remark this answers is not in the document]
RimantasAdeoWeb: Thousands separator

resources/js/src/app/mixins/url.js
RimantasAdeoWeb: Doing it like that would require us to check if the unit is not empty, otherwise spaces would be added needlessly. I think this is a cleaner solution.

resources/js/src/app/mixins/url.js
RimantasAdeoWeb: Not sure if I really understand the problem here. Could you elaborate?

resources/js/src/app/components/itemList/filter/ItemRangeSlider.js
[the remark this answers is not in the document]
TheKeymaster: Makes sense, ok.

resources/js/src/app/mixins/url.js
TheKeymaster: I just rechecked the methods return value, and naming it that would be sufficient. I just got confused with id and name of the actual filter.

resources/js/src/app/mixins/url.js
TheKeymaster: Fine, you win :D!

TheKeymaster (approved):

PR #243, opened 2021-03-05
PLENTY-314 fix website completely breaking down when shop key is not set

RimantasAdeoWeb:


tests/Validators/PluginOrderValidatorTest.php
```diff
@@ -0,0 +1,226 @@
...
+
+        $this->assertEquals($expectedResult, $this->validatorMock->validate());
+    }
+
+    public function validationTestDataProvider(): array
+    {
+        return [
+            'It is valid when IO is loaded before Findologic in the current plugin set' => [
+                [
+                    (object) [
```
TheKeymaster, line 77: Why are these casted into raw objects? You could use a mock, which is created inside of the test instead.

RimantasAdeoWeb: For convenience and saving time while writing these tests. Sure, I could use a mock, but would it be somehow better in any way?

src/Validators/MainValidator.php
```diff
@@ -0,0 +1,91 @@
...
+     * @var LoggerContract
+     */
+    private $logger;
+
+    /**
+     * @return bool
+     */
+    public function validate()
+    {
+        if ($this->validationStatus !== null) {
```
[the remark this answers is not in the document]
RimantasAdeoWeb: Does it? Lame. Do you know how I would get a singleton then?

tests/Validators/PluginOrderValidatorTest.php
RimantasAdeoWeb: Nevermind, found the type. Ok, I'll make these into mocks then

src/Validators/MainValidator.php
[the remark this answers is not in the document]
TheKeymaster: We create two singletons in `\Findologic\Providers\FindologicServiceProvider::register`:

https://github.com/findologic/plugin-plentymarkets-ceres-api/blob/9559159d3706a724bb279c181c334e9b00dfb682/src/Providers/FindologicServiceProvider.php#L23

tests/Validators/PluginOrderValidatorTest.php
TheKeymaster: It is overall easier to change, and I personally prefer mocks over some random objects floating around.

src/Validators/MainValidator.php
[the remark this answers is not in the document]
RimantasAdeoWeb: Yup, I did that as well

TheKeymaster (approved):

PR #245, opened 2021-03-10
Plenty 317

RimantasAdeoWeb:


resources/js/src/app/components/itemList/filter/ItemColorTiles.js
```diff
@@ -19,6 +19,14 @@ Vue.component("item-color-tiles", {
         })
     },
 
+    mounted() {
+        this.$nextTick(function () {
+            $(function() {
```
TheKeymaster, line 24: Do we actually need the `$(function() {...` wrapper, now that we use `mounted`?

resources/views/ItemList/Components/Filter/ItemFilterImage.twig
```diff
@@ -4,7 +4,7 @@
             <div class="form-check" v-for="value in facet.values" :key="value.id">
                 <input :id="'option-' + value.id" class="form-check-input hidden-xs-up" type="checkbox" :checked="value.selected" @change="updateFacet(value)" :disabled="isLoading">
                 <label :for="'option-' + value.id" class="form-check-label" rel="nofollow">
-                    <img :src="value.imageUrl" @error="handleImageError($event)" width="80px"/>
+                    <img :src="value.imageUrl" @error="handleImageError($event)" :width="value.imageUrl ? '80px' : ''"/>
```
TheKeymaster, line 7: How would this look like? Can you add a screenshot of before and after, so we can decide which one we want to use.

resources/views/content/scripts.twig
```diff
@@ -32,12 +32,6 @@
     <script src="{{ plugin_path("Findologic") }}/js/dist/filters-component-min.js"></script>
 {% endif %}
 
-<script>
```
TheKeymaster, line 35: :+1:

TheKeymaster (changes requested):

resources/views/ItemList/Components/Filter/ItemFilterImage.twig
RimantasAdeoWeb: Before empty space would be left after the filter is selected for the image, even when there is no image:
https://snipboard.io/8YVMkm.jpg
After:
https://snipboard.io/3FE1IH.jpg

resources/js/src/app/components/itemList/filter/ItemColorTiles.js
RimantasAdeoWeb: Apparently not, works well without it.

TheKeymaster (approved):

resources/views/ItemList/Components/Filter/ItemFilterImage.twig
TheKeymaster: Thank you, we will leave it as it is.

PR #249, opened 2021-03-25
 PLENTY-318_use_new_method_setDocuments_for_navigation

dblaichinger:


src/Services/FallbackSearchService.php
```diff
@@ -59,7 +53,7 @@ public function handleSearchQuery(Request $request, ExternalSearch $externalSear
      * @param ExternalSearch $externalSearch
      * @return array
      */
-    private function getSearchResults(Request $request, ExternalSearch $externalSearch)
+    public function getSearchResults(Request $request, ExternalSearch $externalSearch)
```
TheKeymaster, line 56: If this is only public for tests, it should also be possible to mock this when you set it to `protected`. This should also be possible for `createResponseFromSearchResult`. I am not 100% certain but I think that these can still be mocked.

tests/Services/SearchServiceTest.php
```diff
@@ -1209,6 +1214,54 @@ public function testInitialNavigationPageGetsResultCountFromPlenty()
...
+        $requestMock = $this->getMockBuilder(HttpRequest::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $externalSearchServiceMock = $this->getMockBuilder(ExternalSearch::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $externalSearchServiceMock->expects($this->once())
+            ->method('setDocuments')
+            ->willThrowException(new Exception('Method setDocuments doesn\t exist'));
```
TheKeymaster, line 1264:
```suggestion
            ->willThrowException(new Exception('Method setDocuments doesn\'t exist'));
```

tests/Services/SearchServiceTest.php
```diff
@@ -1209,6 +1214,54 @@ public function testInitialNavigationPageGetsResultCountFromPlenty()
...
+
+        $this->fallbackSearchService->expects($this->once())
+            ->method('createResponseFromSearchResult')
+            ->willReturn($responseMock);
+
+        $requestMock = $this->getMockBuilder(HttpRequest::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $externalSearchServiceMock = $this->getMockBuilder(ExternalSearch::class)
```
TheKeymaster, line 1259: So using `stdClass` was not possible, could you explain what was the issue?

TheKeymaster (changes requested):

src/Services/FallbackSearchService.php
dblaichinger: It's necessary to call those methods in `SearchService::doNavigation` that's why I set it to public.

tests/Services/SearchServiceTest.php
dblaichinger: Wow that is a really nice feature 👍

tests/Services/SearchServiceTest.php
TheKeymaster: Damn, understandable. Resolved.

dblaichinger:
Closed, renamed branch and new PR is #251

PR #251, opened 2021-04-01
PLENTY-318_increase_navigation_performance_with_ceres_5_0_26

dblaichinger:


plugin.json
```diff
@@ -22,7 +22,7 @@
   ],
   "type": "template",
   "require": {
-    "Ceres": "~5.0",
+    "Ceres": "~5.0.26",
```
TheKeymaster, line 25: Please change this to `>=5.0.26`. This version constraint would limit us to `<5.1`, as it is described here https://getcomposer.org/doc/articles/versions.md#tilde-version-range-.

> The `~` operator is best explained by example: `~1.2` is equivalent to `>=1.2 <2.0.0`, while `~1.2.3` is equivalent to `>=1.2.3 <1.3.0`.

The better way would be to say `>=5.0 <6.0.0`, but I do not think that this constraint is possible, as Plenty is not very good at version parsing in this regard. Maybe it works now but it didn't work in the past AFAIK.

TheKeymaster (changes requested):

plugin.json
dblaichinger: Need to set `>=5.0.26`, the other option doesn't work.
Screen Shot 2021-04-08 at 11 59 54

TheKeymaster (approved):

PR #253, opened 2021-04-09
PLENTY-324 fix default filters not working when an incorrect shopkey …

RimantasAdeoWeb:
…is entered


src/Services/SearchService.php
```diff
@@ -263,10 +268,18 @@ public function search(HttpRequest $request, ExternalSearch $externalSearch)
...
-        $request = $this->requestBuilder->buildAliveRequest();
-        $response = $this->client->call($request);
+        if ($this->aliveTestResult === null) {
+            $request = $this->requestBuilder->buildAliveRequest();
+            $response = $this->client->call($request);
+
+            $this->aliveTestResult = ($response === Plugin::API_ALIVE_RESPONSE_BODY);
+
+            if (!$this->aliveTestResult) {
+                $this->logger->error('Findologic search is not available!');
```
dblaichinger, line 278: @RimantasAdeoWeb 
One point has not been migrated from PLENTY-303 to PLENTY-324, I've done that now. Please update the message and the log level within this PR.

Screen Shot 2021-04-09 at 13 34 24

dblaichinger (changes requested):

RimantasAdeoWeb:
Sorry about the forced push.

src/Services/SearchService.php
RimantasAdeoWeb: Moved the logging back to the middleware, where it was originally.

resources/js/src/app/components/itemList/filter/ItemFilter.js
```diff
@@ -31,6 +37,7 @@ Vue.component("findologic-item-filter", {
 
     created() {
         this.$options.template = this.template || "#vue-item-filter";
+        this.facetType = (typeof this.facet.findologicFilterType !== 'undefined') ? this.facet.findologicFilterType : this.facet.type;
```
TheKeymaster, line 40: When is this `undefined`? Is this the case when Findologic is not active?

src/Middlewares/Middleware.php
```diff
@@ -76,14 +79,19 @@ public function before(Request $request)
         }
 
         if (!$this->searchService->aliveTest()) {
-            $this->getLoggerObject()->error('Findologic search is not available!');
-
+            $this->getLoggerObject()->alert(
```
TheKeymaster, line 82: So we had some internal discussion about this and decided to actually get rid of this log line.

Reason behind that is that we do not add such a log line for other plugins as well. Also a log line should inform someone to take action. In that case the customer can not take any action, as this could be a one time accident and if Findologic was actually down, the customer would notice that anyway and inform us via the Plenty forum or via our support.

src/Middlewares/Middleware.php
```diff
@@ -76,14 +79,19 @@ public function before(Request $request)
...
+                'Findologic search is not available! Using Plentymarkets search/navigation for this request.'
+            );
             return;
         }
 
         $this->isSearchPage = strpos($request->getUri(), '/search') !== false;
         $this->activeOnCatPage = !$this->isSearchPage && $this->pluginConfig->get(Plugin::CONFIG_NAVIGATION_ENABLED);
 
+        if (!$this->isSearchPage && !$this->activeOnCatPage) {
```
TheKeymaster, line 91: Defensive Coding :raised_hands:

src/Services/SearchService.php
```diff
@@ -77,6 +77,11 @@ class SearchService implements SearchServiceInterface
      */
     protected $configRepository;
 
+    /**
+     * @var bool
+     */
+    protected $aliveTestResult;
```
TheKeymaster, line 83: Is this somehow cached now? Also the typehint is kinda misleading, it should actually be `bool|null`.

TheKeymaster (changes requested):

resources/js/src/app/components/itemList/filter/ItemFilter.js
RimantasAdeoWeb: If I remember correctly, this happens when the shopkey is correct but Findologic navigation filters are disabled.

src/Services/SearchService.php
RimantasAdeoWeb: Yeah, it's a leftover from a previous attempt at fixing this. It's not really used now, but I thought it could be left this way as it might be useful in the future.

resources/js/src/app/components/itemList/filter/ItemFilter.js
TheKeymaster: When the Findologic navigation filters are disabled, the Plentymarkets filter should be shown, is this the case and can these be properly selected? If this is the case, this comment can be resolved, otherwise please investigate.

src/Services/SearchService.php
TheKeymaster: Okay, but please update the docblock.

TheKeymaster (approved):

PR #257, opened 2021-04-21
PLENTY-309 do not show filter frequency for selected values

RimantasAdeoWeb:


resources/views/ItemList/Components/Filter/ItemFilter.twig
```diff
@@ -61,7 +61,7 @@
             <div class="form-check" v-for="value in facet.values" :key="value.id">
                 <input :id="'option-' + value.id" class="form-check-input hidden-xs-up" type="checkbox" :checked="value.selected" @change="updateFacet(value)" :disabled="isLoading">
                 <label :for="'option-' + value.id" class="form-check-label" v-text="value.name" rel="nofollow"></label>
-                <div class="filter-badge" v-text="value.count"></div>
+                <div v-if="!value.selected" class="filter-badge" v-text="value.count"></div>
```
TheKeymaster, line 64: Does this work for all filter types, also including vendor image, or others?

TheKeymaster (changes requested):

resources/views/ItemList/Components/Filter/ItemFilter.twig
RimantasAdeoWeb: Good call, it wasn't working with images. All the others look fine.

resources/views/ItemList/Components/Filter/ItemFilterImage.twig
```diff
@@ -7,7 +7,7 @@
                     <img :src="value.imageUrl" @error="handleImageError($event)" :width="value.imageUrl ? '80px' : ''"/>
                     <span v-text="value.name"></span>
                 </label>
-                <div class="filter-badge" v-if="value.count" v-text="value.count"></div>
+                <div class="filter-badge" v-if="!value.selected" v-text="value.count"></div>
```
TheKeymaster, line 10: Shouldn't this be `value.count && !value.selected`? Otherwise this still may be shown, even if there is no count which could cause problems.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #259, opened 2021-04-29
PLENTY-310 show number of selected filter values

RimantasAdeoWeb:


resources/js/src/app/components/itemList/filter/ItemFilter.js
```diff
@@ -37,6 +37,18 @@ Vue.component("findologic-item-filter", {
             selectedFacets: state => state.itemList.selectedFacets,
             isLoading: state => state.itemList.isLoading
         }),
+
+        selectedValuesCount() {
```
TheKeymaster, line 41: I prefer having this as a computed property.

resources/js/src/app/components/itemList/filter/ItemFilter.js
```diff
@@ -37,6 +37,18 @@ Vue.component("findologic-item-filter", {
...
+            let count = 0;
+
+            this.facet.values.forEach(function(value) {
```
TheKeymaster, line 44: How about this? It should yield the same result (I didn't test it), but is easier to read in my opinion.

```js
const selectedFacets = this.facet.values.filter(function (value) {
    return value.selected;
});

return selectedFacets.length;
```

TheKeymaster: Also don't we already have a helper `getSelectedFilters` for this purpose, why not reuse the existing logic? This way we already could utilize `getSelectedFilters().length` in the template.

TheKeymaster (changes requested):

resources/js/src/app/components/itemList/filter/ItemFilter.js
RimantasAdeoWeb: But it IS a computed property. Or am I misunderstanding something?

resources/js/src/app/components/itemList/filter/ItemFilter.js
RimantasAdeoWeb: getSelectedFilters returns the selected filters, not the selected values for each filter. So extra stuff would have to be done which would end up being messier than this, unfortunately.

TheKeymaster: Ah I see.

resources/js/src/app/components/itemList/filter/ItemFilter.js
TheKeymaster: Oh, it already is. Then forget this :)

TheKeymaster (approved):

PR #264, opened 2021-05-12
PLENTY-327 send ceres version with search requests

RimantasAdeoWeb:


src/Services/PluginInfoService.php
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * @var PluginRepositoryContract|null
+     */
+    private $pluginRepository;
+
+    /**
+     * @var CachingRepository
+     */
+    private $cache;
```
TheKeymaster, line 22: If this service is registered as singleton, you should be able to get these via constructor injection out-of-the-box.

src/Services/PluginInfoService.php
```diff
@@ -0,0 +1,90 @@
+<?php
+
+namespace Findologic\Services;
+
+use Plenty\Modules\Plugin\Contracts\PluginRepositoryContract;
+use Plenty\Modules\Plugin\Models\Plugin;
+use Plenty\Modules\Plugin\PluginSet\Contracts\PluginSetRepositoryContract;
+use Plenty\Plugin\CachingRepository;
+
+class PluginInfoService
```
TheKeymaster, line 10: Good idea moving this to a separate file.

src/Services/PluginInfoService.php
```diff
@@ -0,0 +1,90 @@
...
+        $cachedVersion = $this->getCache()->get(self::PLUGIN_VERSION_CACHE_KEY_PREFIX . $pluginName);
+        if ($cachedVersion !== null) {
+            return $cachedVersion;
+        }
+
+        if (!$plugin = $this->getPlugin($pluginName)) {
+            return null;
+        }
+
+        if (!$plugin->versionProductive || $plugin->versionProductive == '0.0.0') {
```
TheKeymaster, line 43: Maybe put `0.0.0` into a separate constant e.g. `DEFAULT_VERSION` or similar. Maybe you can come off with a better name.

tests/Services/SearchServiceTest.php
```diff
@@ -1121,17 +1132,7 @@ public function redirectToProductPageOnDoSearchProvider(): array
...
-            ->disableOriginalConstructor()
-            ->setMethods([])
-            ->getMock();
-        $searchServiceMock->method('getPluginRepository')->willReturn($pluginRepositoryMock);
-        $pluginRepositoryMock->method('getPluginByName')->willReturn($pluginMock);
-        $pluginMock->versionProductive = $version;
+        $this->pluginInfoService->method('getPluginVersion')->willReturn($version);
 
         $this->assertEquals($shouldFilter, $searchServiceMock->shouldFilterInvalidProducts());
     }
```
TheKeymaster, line 1138: Seems like I can not add any comments to files, which have been added but do not have any contents. It seems like you have commited an empty file called `versionProductive`. Do we need this or was this commited by accident?

src/Services/PluginInfoService.php
```diff
@@ -0,0 +1,90 @@
...
+     */
+    private $pluginRepository;
+
+    /**
+     * @var CachingRepository
+     */
+    private $cache;
+
+    /**
+     * @var Plugin|null[]
```
TheKeymaster, line 25: That typehint is either wrong or it is used in a wrong way. It is supposed to store an array of plugins, but the typehint either contains a single plugin or an array of null, which doesn't really make sense. I think the proper typehint should be `Plugin[]`.

TheKeymaster (changes requested):

src/Services/PluginInfoService.php
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * @return Plugin|null
+     */
+    private function getPlugin(string $pluginName)
+    {
+        if (!isset($this->plugins[$pluginName])) {
+            $this->plugins[$pluginName] = $this->getPluginRepository()->getPluginByName($pluginName);
+        }
+
+        return $this->plugins[$pluginName];
```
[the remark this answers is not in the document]
RimantasAdeoWeb: The story description asked for a separate service for this ("Centralize this code, to avoid duplication."). I thought I would make a service for all things having to do with plugin info, not just for retrieving the version. So in the future if more methods would be added here to get different info about the plugin, it wouldn't need to reload the plugin for each bit of info it needs to retrieve.

src/Services/PluginInfoService.php
```diff
@@ -0,0 +1,90 @@
...
+     */
+    private $cache;
+
+    /**
+     * @var Plugin|null[]
+     */
+    private $plugins = [];
+
+    /**
+     * @return string|null
```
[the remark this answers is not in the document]
RimantasAdeoWeb: Unfortunately Plentymarkets runs on an ancient version of PHP which doesn't support nullable returns.

src/Services/PluginInfoService.php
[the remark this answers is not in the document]
TheKeymaster: Ok makes sense.

TheKeymaster (approved):

PR #274, opened 2021-07-12
PLENTY-346 add minimum search term length config

RimantasAdeoWeb:


resources/js/src/app/components/itemList/ItemSearch.js
```diff
@@ -59,13 +59,10 @@ Vue.component("item-search", {
 
     methods:
     {
-        prepareSearch()
+        search()
```
TheKeymaster, line 62: Why was this changed?

resources/views/content/scripts.twig
```diff
@@ -1,28 +1,23 @@
...
-            action = '/' + App.language + '/search';
-        }
-
-        if (formElement.querySelector('input.search-input').value.trim() !== '') {
-            handleSearchQuery(formElement, action);
-        } else {
-            handleSmartSuggestClick(formElement, action);
-        }
-    }
+                return;
```
TheKeymaster, line 13: Good early return :+1:

resources/views/content/scripts.twig
```diff
@@ -56,21 +51,23 @@
...
+    <form method="GET" action="/search" @submit="search()">
         <div class="container-max">
             <div class="position-relative">
                 <div class="d-flex flex-grow-1 position-relative my-2">
                     <input name="query"
                            type="search"
                            class="search-input flex-grow-1 px-3 py-2"
                            ref="searchInput"
-                           @keyup.enter="prepareSearch()"
+                           {% if minSearchLength %}
```
TheKeymaster, line 62: Is it possible that this does not exist?

src/Components/PluginConfig.php
```diff
@@ -151,6 +151,18 @@ public function getShopkeys()
         return $this->shopkeys;
     }
 
+    /**
+     * return int
```
TheKeymaster, line 155: This may be removed and properly typehinted via `: int`.

src/Components/PluginConfig.php
```diff
@@ -151,6 +151,18 @@ public function getShopkeys()
...
+     */
+    public function getMinSearchLength()
+    {
+        $minSearchLength = $this->configRepository->get(Plugin::CONFIG_MIN_SEARCH_LENGTH, 0);
+
+        $isPositiveInteger = (preg_match('/^\d+$/', $minSearchLength) === 1);
```
TheKeymaster, line 161: Creative way of using a regex :+1:.

config.json
```diff
@@ -27,6 +27,14 @@
               }
             ]
           }
+        },
+        "minSearchLength": {
```
TheKeymaster, line 31: Can we rename this to `minimalSearchTermLength`? If it is possible always be as descriptive as possible without removing parts that may be important to know. This only applies if it makes sense and shouldn't be done for naming that is already common sense.

That one is more of a gut feeling though, as "search" could be synonymous with "query" or "search term", so I am up for an argument, in case you want to leave it as it is.

resources/views/content/scripts.twig
TheKeymaster: Maybe you want to change this to `> 0`, so it is clear that this will always be a number?

TheKeymaster (changes requested):

resources/js/src/app/components/itemList/ItemSearch.js
RimantasAdeoWeb: Can't remember exactly. Either prepareSearch() was no longer needed after refactoring or it was actually causing unwanted behavior.

RimantasAdeoWeb: It relates to the changes in scripts.twig. There was code in two places that did essentially the same thing. I refactored it and the separate prepareSearch() function was no longer needed.

TheKeymaster (approved):

PR #276, opened 2021-08-11
PLENTY-356 handle curl errors

RimantasAdeoWeb:


tests/Api/Response/ResponseParserTest.php
```diff
@@ -79,6 +79,46 @@ public function testParse()
...
+    public function testHandleErrorResponse()
+    {
+        $errorResponse = [
+            'error' => true,
+            'error_no' => 0,
+            'error_msg' => 'Curl error: Could not resolve host: service.findologic.com',
+            'error_file' => '/findologic/http_request2/HTTP/Request2/Adapter/Curl.php',
+            'error_line' => 155
+        ];
+        $expectedExceptionMessage = 'simplexml_load_string() expects parameter 1 to be string, array given';
```
TheKeymaster, line 91: We shouldn't even get to the point where `simplexml_load_string` tries to parse an array. We must catch this before and handle this case properly, without relying that passing data that can not be parsed by the parser, throwing an exception. Issues that we know beforehand that can be catched before an actual exception occurs should always be preferred.

For this please check if the data coming is an actual array and in case it is, log an appropriate error message e.g. `Response could not be parsed`, maybe even add the response as a log context.

Also since there is no exception, we do not need to log this.

TheKeymaster (changes requested):

src/Api/Response/ResponseParser.php
```diff
@@ -34,11 +34,17 @@ public function __construct(
         $this->logger = $loggerFactory->getLogger(Plugin::PLUGIN_NAMESPACE, Plugin::PLUGIN_IDENTIFIER);
     }
 
-    public function parse(HttpRequest $request, string $responseData): Response
+    public function parse(HttpRequest $request, $responseData): Response
     {
         /** @var Response $response */
         $response = $this->createResponseObject();
 
+        if (!is_string($responseData)) {
```
TheKeymaster, line 42: Cool! That was exactly what I meant :+1:

TheKeymaster (approved):

PR #285, opened 2021-09-06
PLENTY-373_use_correct_language_path

dblaichinger:


tests/Middlewares/MiddlewareTest.php
```diff
@@ -369,6 +370,71 @@ public function testFindologicIsTriggeredIfProperlyConfigured(
...
+            ->willReturn($webstoreConfigMock);
+
+        $localizationRepositoryMock = $this->getMockBuilder(LocalizationRepositoryContract::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $localizationRepositoryMock->expects($this->once())
+            ->method('getLanguage')
+            ->willReturn($usedLanguage);
+
+        global $classInstances;
```
TheKeymaster, line 415: Can we reset the state of this in the `tearDown` to just be an empty array after the test has run?

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #287, opened 2021-10-15
PLENTY-383 making ceres pdp redirect compatible with new product id structure

msilinskas:


src/Services/SearchService.php
```diff
@@ -377,6 +377,11 @@ private function getProductDetailUrl(Response $response)
         $itemSearchService = $this->getItemSearchService();
 
         $productId = $response->getProductsIds()[0];
+
+        if (strpos($productId, "_")) {
```
TheKeymaster, line 381:
```suggestion
        if (strpos($productId, '_')) {
```

PR #290, opened 2021-12-20
PLENTY-412_fix_landingpage_redirect

TheKeymaster:


src/Services/SearchService.php
```diff
@@ -154,7 +154,7 @@ public function doSearch(HttpRequest $request, ExternalSearch $externalSearch)
     {
         $results = $this->search($request, $externalSearch);
 
-        if ($results->getResultsCount() == 0) {
+        if ($results->getResultsCount() == 0 && !$results->getLandingPage()) {
```
TheKeymaster, line 157: This little thing took me ages to figure out.. Otherwise this would've been a 2min fix.

dblaichinger: As result count and landingpage is independent, how about adding the redirect to landingpage above this one? Landingpage has priority anyway.

dblaichinger (changes requested):

PR #291, opened 2021-12-20
PLENTY-411_retry_mechanism

dblaichinger:


src/Services/SearchService.php
```diff
@@ -472,4 +475,30 @@ private function buildItemURL($itemData, $withVariationId = true): string
...
         }
     }
+
+    /**
+     * @param Request $request
+     * @return mixed
+     */
+    private function requestWithRetry(Request $request)
+    {
+        for ($i = 0; $i <= self::MAX_RETRIES; $i++) {
```
TheKeymaster, line 485: A do while loop would simplify the logic.

```php
do {
    // Send request...

    if ($successful) {
        return $response;
    }

    // Logging...
    $i++;
} while ($i <= self::MAX_RETRIES)
```

src/Services/SearchService.php
```diff
@@ -472,4 +475,30 @@ private function buildItemURL($itemData, $withVariationId = true): string
...
+            $errorMsg = '';
+            $responseData = $this->client->call($request);
+            
+            if (is_array($responseData) && array_key_exists('error', $responseData)) {
+                $errorMsg = 'Plentymarkets returned error response';
```
TheKeymaster, line 490:
```suggestion
                $errorMsg = 'Plentymarkets SDK returned an error response';
```

tests/Api/Response/ResponseParserTest.php
```diff
@@ -94,11 +94,6 @@ public function testHandleErrorResponse()
         /** @var Request|MockObject $requestMock */
         $requestMock = $this->getMockBuilder(Request::class)->disableOriginalConstructor()->setMethods([])->getMock();
 
-        $this->logger
```
TheKeymaster, line 97: Shouldn't this now ensure that the retry is logged?

tests/Services/SearchServiceTest.php
```diff
@@ -1283,6 +1283,45 @@ public function testInitialNavigationPageGetsResultCountFromPlenty()
...
+            'error_file' => '/var/www/SdkRestApi.php',
+        ];
+        $nonStringErrorResponse = false;
+        $validResponse = $this->getMockResponse('someResultsWithFilters.xml');
+
+        $this->client->expects($this->exactly(3))
+            ->method('call')
+            ->willReturnOnConsecutiveCalls($plentyErrorResponse, $nonStringErrorResponse, $validResponse);
+
+        $this->logger->expects($this->at(0))->method('error')->with(
```
TheKeymaster, line 1312: `at()` [is deprecated](https://github.com/sebastianbergmann/phpunit/issues/4297). Use `exactly()` and `withConsecutive()` instead. Similar how you already did it in line `1308`.

src/Services/SearchService.php
```diff
@@ -472,4 +475,30 @@ private function buildItemURL($itemData, $withVariationId = true): string
             )->toRelativeUrl($includeLanguage);
         }
     }
+
+    /**
+     * @param Request $request
+     * @return mixed
+     */
+    private function requestWithRetry(Request $request)
```
TheKeymaster, line 483:
```suggestion
    private function requestWithRetries(Request $request)
```

src/Services/SearchService.php
```diff
@@ -472,4 +475,30 @@ private function buildItemURL($itemData, $withVariationId = true): string
             )->toRelativeUrl($includeLanguage);
         }
     }
+
+    /**
+     * @param Request $request
```
TheKeymaster, line 480: Only add things to the docblock, that aren't already part of the PHP native typings. `mixed` is fine, as it can not be typehinted with PHP 7.0, but `Request` is already typed by PHP.

src/Services/SearchService.php
```diff
@@ -472,4 +475,30 @@ private function buildItemURL($itemData, $withVariationId = true): string
...
+    {
+        for ($i = 0; $i <= self::MAX_RETRIES; $i++) {
+            $errorMsg = '';
+            $responseData = $this->client->call($request);
+            
+            if (is_array($responseData) && array_key_exists('error', $responseData)) {
+                $errorMsg = 'Plentymarkets returned error response';
+            } elseif (!is_string($responseData)) {
+                $errorMsg = 'Invalid response received from server';
+            }
```
TheKeymaster, line 493: Wouldn't it make sense to move this to a separate method? E.g. `getErrorMessage`. This way we do not need to default to an empty string, which confused me when I initially read this, until I saw that we do an empty check at the end. This way we could just do a null check and `getErrorMessage` only returns an error if there is one, otherwise `null`.

.gitignore
```diff
@@ -4,3 +4,4 @@ build/
 .idea/
 src/index.php
 src/Mocks
+tests/build
```
TheKeymaster, line 7: Why was this added? When is this directory created? It does not seem to be related to your change.

tests/Services/SearchServiceTest.php
```diff
@@ -1283,6 +1283,45 @@ public function testInitialNavigationPageGetsResultCountFromPlenty()
...
+        $externalSearchServiceMock = $this->getMockBuilder(ExternalSearch::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $externalSearchServiceMock->expects($this->never())->method('setResults');
+
+        $this->requestBuilder->expects($this->once())->method('build')
+            ->willReturn(new Request());
+
+        $plentyErrorResponse = [
+            'error' => true,
```
TheKeymaster, line 1301: If I recall correctly there is also an `error_code` which was set to `404`.

src/Services/SearchService.php
```diff
@@ -33,6 +34,7 @@ class SearchService implements SearchServiceInterface
     use Loggable;
 
     const DEFAULT_ITEMS_PER_PAGE = 25;
+    const MAX_RETRIES = 2;
```
TheKeymaster, line 37: Can we rename this to `MAX_TRIES` and change its value to `3` and make sure exactly `3` tries are being made? Until I saw the test I was assuming that we only do 2 tries.

TheKeymaster (changes requested):

src/Services/SearchService.php
TheKeymaster: I'm fine with this. I was just wondering if we could improve this somehow, to make it more clear that actually 3 requests are being made in total, but I guess we really need some refactoring for this in the future, as the current state isn't in its best shape.

TheKeymaster (approved):

PR #292, opened 2022-01-04
PLENTY-414_PH-456_log_xml_parse_error

dblaichinger:


tests/Api/Response/ResponseParserTest.php
```diff
@@ -73,16 +73,46 @@ public function testParse()
...
+                    ['xmlString' => 'invalid-xml'],
+                ]
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider responseDataProvider
+     */
+    public function testHandleInvalidResponse($response, $expectedLog)
```
TheKeymaster, line 114: Add some typehints to the arguments. It is also not clear what type `$expectedLog` is, by the name. Maybe rename it to `$expectedLogData`, so it is more clear what it contains.

```suggestion
    /**
     * @param string|array $response
     */
    public function testHandleInvalidResponse($response, array $expectedLog)
```

TheKeymaster: Maybe also split up `$errorMessage` and `$context` into separate variables. This way you don't have to access the array indexes in line 127.

TheKeymaster (changes requested):

TheKeymaster (approved):

PR #294, opened 2022-01-12
PLENTY-415_fix_count_in_retry_mechanism

dblaichinger:


src/Services/SearchService.php
```diff
@@ -531,8 +532,8 @@ private function validateResponse($responseData)
         if (is_array($responseData) && array_key_exists('error', $responseData) && $responseData['error'] === true) {
             $errorMsg = 'Plentymarkets SDK returned an error response';
         } elseif (!is_string($responseData)) {
-            $errorMsg = 'Invalid response received from server';
-        } elseif ($responseData === '') {
+            $errorMsg = 'Plentymarkets SDK returned invalid response - Expected string';
```
TheKeymaster, line 535: Maybe to make this more clear, by logging what the type actually was.
```suggestion
            $errorMsg = sprintf('Plentymarkets SDK returned invalid response (expected string, received %s)', gettype($responseData));
```

TheKeymaster (changes requested):

src/Services/SearchService.php
dblaichinger: Tried `gettype` before and guess what, Plenty doesn't allow it...

TheKeymaster: Great. Nvm then

TheKeymaster (approved):

PR #298, opened 2022-02-08
Plenty 420 redirection url dependence on cheapest variation

msilinskas:


tests/Services/SearchServiceTest.php
```diff
@@ -332,154 +337,893 @@ public function categorySearchProvider(): array
...
+                ],
+                'redirectUrl' => '/test-product_11',
+                'attributes' => [
+                    'page' => 1
+                ],
+                'de',
+                'de',
+                false
+            ],
+            'One product with three variations, main variation without price, no query matches' => [
```
TheKeymaster, line 909: Since almost all tests create three similar variation, maybe you can combine the method from above with an additional helper that creates a set amount of variations, based on data for each variation.

TheKeymaster (changes requested):

tests/Services/SearchServiceTest.php
```diff
@@ -956,6 +1760,11 @@ public function redirectToProductPageOnDoSearchProvider(): array
                                 'barcodes' => [
                                     ['code' => '123123123'],
                                     ['code' => '321321321']
+                                ],
+                                'salesPrices' => [
```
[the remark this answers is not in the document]
msilinskas: Yes, the test fails then, because we've added a price condition in the \Findologic\Services\SearchService::getVariationIdForRedirect method. I can't reproduce the way, that item doesn't have a salesPrice set.

src/Services/SearchService.php
```diff
@@ -506,6 +517,26 @@ private function getVariationIdForRedirect(string $query, array $documents)
         return $cheapestVariationId;
     }
 
+    /**
+     * @return float|null
```
TheKeymaster, line 521: How can this return `null`? Isn't the default price a `float`? `float` can be natively typed.

src/Services/SearchService.php
```diff
@@ -427,20 +427,31 @@ private function getProductDetailUrl(Response $response)
...
         }
 
-        $withVariationId = false;
+        $withVariationId = $this->shouldExportWithVariationId($variationId);
+
+        return $this->buildItemURL($productData, $withVariationId);
+    }
+
+    /**
+     * @return bool
```
TheKeymaster, line 436: Why can't this be natively typed?

tests/Services/SearchServiceTest.php
```diff
@@ -501,83 +542,25 @@ public function redirectToProductPageOnDoSearchProvider(): array
                 'variationSearchByItemIdResult' => [
                     'success' => true,
                     'total' => 2,
-                    'documents' => [
+                    'documents' => $this->getMultipleItemsDocuments(
```
TheKeymaster, line 545: Wow! Look at this cut-down. So much more pleasing to read. Nice :+1:

TheKeymaster (approved):

PR #303, opened 2022-02-21
PLENTY-427 category filter on category pages fix

msilinskas:


resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.js
```diff
@@ -19,12 +24,24 @@ Vue.component("item-category-dropdown", {
...
+            return typeof this.getSelectedFilters().find(element => element.id == this.facet.id) !== 'undefined';
+        },
     },
 
     methods: {
         getSubCategoryName(parentCategory, subCategory) {
             return parentCategory.name + '_' + subCategory.name;
+        },
+
+        categorySelected(category) {
```
TheKeymaster, line 43: Why would you declare this method? It is the same as `isCategorySelected`.

resources/js/src/app/mixins/url.js
```diff
@@ -277,6 +277,27 @@ export default {
             return selectedFilters;
         },
 
+        /**
+         * @param category
```
TheKeymaster, line 281: Please add a typehint.

resources/js/src/app/mixins/url.js
```diff
@@ -277,6 +277,27 @@ export default {
...
+         * @returns {boolean}
+         */
+        isCategorySelected(category) {
+            let selectedFilters = this.getSelectedFilters(),
+                splitedSelectedCategories = [];
```
TheKeymaster, line 286: Define variables separately on a separate line from eachother.

resources/js/src/app/mixins/url.js
```diff
@@ -277,6 +277,27 @@ export default {
             return selectedFilters;
         },
 
+        /**
+         * @param category
+         * @returns {boolean}
+         */
+        isCategorySelected(category) {
+            let selectedFilters = this.getSelectedFilters(),
```
TheKeymaster, line 285: This could be a `const`.

resources/js/src/app/mixins/url.js
```diff
@@ -277,6 +277,27 @@ export default {
...
+                splitedSelectedCategories = [];
+
+            for (let i = 0; i < selectedFilters.length; i++) {
+                if (selectedFilters[i].id !== this.facet.id) {
+                    continue;
+                }
+
+                splitedSelectedCategories = selectedFilters[i].name.split('>');
```
TheKeymaster, line 293: Wouldn't this cause a conflict when multiple categories have the same name? E.g.

* Men > Shoes
* Women > Shoes

tests/Api/Request/ParametersBuilderTest.php
```diff
@@ -83,7 +83,35 @@ public function setSearchParamsProvider()
                         0 => 'variation_id'
                     ],
                     'attrib' => [
-                        'color' => ['red', 'blue']
+                        'color' => ['red', 'blue'],
+                        'cat' => ['Category 2']
```
TheKeymaster, line 87: Why was this added? The other parameters make sense, but why do we need a category now?

tests/Api/Request/ParametersBuilderTest.php
```diff
@@ -83,7 +83,35 @@ public function setSearchParamsProvider()
...
+                    ],
+                    'selected' => ['cat' => ['Category']],
+                    'order' => 'price ASC',
+                    'count' => 0,
+                    'first' => 0
+                ]
+            ],
+            'Category page request with subcategory selected' => [
+                'parameters' => [
+                    Plugin::API_PARAMETER_ATTRIBUTES => [
```
TheKeymaster, line 97: Isn't `attrib` as a string used everywhere else? I typically prefer to not use constants in unit-tests, as if the constant changes in the future, you can see which components would be affected.

resources/js/src/app/components/itemList/filter/ItemFilter.js
```diff
@@ -63,5 +67,9 @@ Vue.component("findologic-item-filter", {
                 name: parentCategory.name + '_' + subCategory.name
             };
         },
+
+        categorySelected(category) {
```
TheKeymaster, line 71: Please add a return typehint.

resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.js
TheKeymaster, line 43: Please add a return typehint.

resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.js
```diff
@@ -19,12 +24,24 @@ Vue.component("item-category-dropdown", {
             }
 
             return label;
-        }
+        },
+
+        isSelected() {
```
TheKeymaster, line 29: Please add a return typehint.

resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.js
```diff
@@ -19,12 +24,24 @@ Vue.component("item-category-dropdown", {
...
+            if (this.facet.id === 'cat' && this.facet.values.length === 1) {
+                return true;
+            }
+
+            return typeof this.getSelectedFilters().find(element => element.id == this.facet.id) !== 'undefined';
```
TheKeymaster, line 34: Why aren't you using strict comparison when checking the id?
```suggestion
            return typeof this.getSelectedFilters().find(element => element.id === this.facet.id) !== 'undefined';
```

TheKeymaster (changes requested):

resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.js
```diff
@@ -7,6 +7,11 @@ Vue.component("item-category-dropdown", {
     computed: {
         dropdownLabel() {
             let selectedFilters = this.getSelectedFilters();
+
+            if (selectedFilters.length === 0 && this.facet.id === 'cat' && this.facet.values.length === 1) {
```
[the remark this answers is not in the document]
msilinskas: We need this check for the default label setting in the category page.

tests/Api/Request/ParametersBuilderTest.php
msilinskas: We have commented out the continue in the [src/Api/Request/ParametersBuilder](https://github.com/findologic/plugin-plentymarkets-ceres-api/pull/303/files#diff-e93f58cf887da50267647fd05d8598121210af7bb2924259e16341590c69d567)::87. Now we are setting cat attribute to the request, so I had to add it here too.

resources/js/src/app/components/itemList/filter/ItemFilter.js
```diff
@@ -63,5 +69,19 @@ Vue.component("findologic-item-filter", {
...
+        /**
+         * @returns {boolean}
+         */
+        shouldShowFilter() {
+            return this.facet.id !== 'cat' ||
+                this.facet.id === 'cat' && typeof this.showCategoryFilter === 'undefined' ||
+                this.facet.id === 'cat' && this.showCategoryFilter;
+        },
+
+        shouldShowCategoryFilter() {
```
TheKeymaster, line 76: Great that you also added typehints to most methods. Here you could also add one :)

src/Containers/SearchFilterContainer.php
```diff
@@ -19,12 +21,20 @@ public function call(Twig $twig, SearchService $searchService): string
         }
 
         $searchResults = $searchService->getResults();
+        $currentCategory = $searchService->getCategoryService()->getCurrentCategory();
+        $showCategoryFilter = true;
+
+        if (null !== $currentCategory && $currentCategory->level !== 1) {
```
TheKeymaster: Please don't use [yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions).

```suggestion
        if ($currentCategory !== null && $currentCategory->level !== 1) {
```

TheKeymaster: Maybe you wanna change the condition to `> 1` to make it more clear that everything above should not be shown. Also a comment might be helpful, otherwise someone reading this may be confused, why the category filter is not shown further levels below.

resources/js/src/app/mixins/url.js
TheKeymaster: Is this still an issue? If not, please resolve the conversation.

resources/js/src/app/components/itemList/filter/ItemFilter.js
```diff
@@ -63,5 +69,19 @@ Vue.component("findologic-item-filter", {
                 name: parentCategory.name + '_' + subCategory.name
             };
         },
+
+        /**
+         * @returns {boolean}
+         */
+        shouldShowFilter() {
```
TheKeymaster: Can't `shouldShowFilter` and `shouldShowCategoryFilter` be combined somehow?

TheKeymaster (changes requested):

resources/js/src/app/components/itemList/filter/ItemFilter.js
msilinskas: I've moved the checking if facet id is equals to cat to the twig file and left only shouldShowCategoryFilter method in the js component.

TheKeymaster (approved):

PR #305, opened 2022-02-22
PLENTY-428_review_3_8_0

dblaichinger:


src/Services/SearchService.php
```diff
@@ -434,7 +435,9 @@ private function getProductDetailUrl(Response $response)
 
     private function shouldExportWithVariationId(int $variationId): bool
     {
-        $showPleaseSelect = $this->pluginInfoService->isOptionShowPleaseSelectEnabled('Ceres');
+        /** @var TemplateConfigService $templateConfigService */
+        $templateConfigService = pluginApp(TemplateConfigService::class);
+        $showPleaseSelect = $templateConfigService->getBoolean('item.show_please_select');
```
TheKeymaster, line 440: I still think it makes sense to have the method, as it hides the internal naming schema and exposes plugin configuration in an easily obtainable way. The `PluginInfoService` could also have the `TemplateConfigService` available inside the constructor to load it automatically via dependency injection.

tests/Services/SearchServiceTest.php
```diff
@@ -121,6 +127,13 @@ public function setUp()
             ->disableOriginalConstructor()
             ->setMethods([])
             ->getMock();
+        $this->templateConfigService = $this->getMockBuilder(TemplateConfigService::class)
+            ->disableOriginalConstructor()
+            ->setMethods([])
+            ->getMock();
+
+        global $classInstances;
```
TheKeymaster, line 135: I read `global` and my brain goes brrr. Sad though that we need to opt for this way, as we're really restricted by PHPUnit and Plentymarkets here.

tests/Services/SearchServiceTest.php
```diff
@@ -713,7 +727,7 @@ public function redirectToProductPageOnDoSearchProvider(): array
                 ],
                 'language' => 'de',
                 'defaultLanguage' => 'de',
-                false
+                'optionShowPleaseSelect' => false
```
TheKeymaster, line 730: :pray: thanks for adding the name of the parameter

tests/Services/SearchServiceTest.php
```diff
@@ -229,7 +242,7 @@ public function testRedirectToProductPageOnDoSearch(
         array $attributes,
         string $language = 'de',
         string $defaultLanguage = 'de',
-        bool $isOptionShowPleaseSelectEnabled = true
+        bool $optionShowPleaseSelect = true
```
TheKeymaster, line 245: I think I mentioned this in the last review with Modestas. The thing is the naming itself "optionShowPleaseSelect" should be renamed, as a boolean should typically make sure to expose his state by making it clear what happens if that state is reached. Something like `isEnabled` or `hasSticker` are pretty obvious namings, as they refer directly to the state and it is easy to understand what happens if their state changes from `true` to `false` or vice versa. 

I am fine with this for now, but we should definitely keep an eye out for that in the future.

resources/js/src/app/components/itemList/filter/ItemColorTiles.js
```diff
@@ -1,6 +1,5 @@
 import url from '../../../mixins/url'
 import { SVGInjector } from '@tanem/svg-injector'
-import DOMPurify from 'dompurify'
```
TheKeymaster, line 3: Seems like the compiler didn't care after all, and didn't even include this in the first place. In that case no customer was affected by this anyway. Although you already indicated this by click-testing, the fact that the bundle itself wasn't changed confirmed that.

TheKeymaster (changes requested):

src/Services/PluginInfoService.php
```diff
@@ -70,43 +75,11 @@ public function getPluginVersion(string $pluginName)
...
-        }
-
-        $this->cache->put(
-            self::OPTION_SHOW_PLEASE_SELECT_CACHE_KEY_PREFIX . $pluginName,
-            $value,
-            60*24
-        );
-
-        return $value;
+        return $this->templateConfigService->getBoolean('item.show_please_select');
```
TheKeymaster, line 82: Boom! As easy as that. So nice! :+1:

TheKeymaster (approved):

PR #306, opened 2022-03-03
PLENTY-429_plentyShop_5_0_47_compatibility

dblaichinger:
PLENTY-429_plentyShop_5_0_47_compatibility

### DIFF
--- a/src/Services/SearchService.php
+++ b/src/Services/SearchService.php
@@ -421,15 +421,14 @@ class SearchService implements SearchServiceInterface
         // Necessary to retrieve variation data and prices
         $resultFields = $this->loadResultFieldTemplate(ResultFieldTemplate::TEMPLATE_LIST_ITEM);
 
-        $variationSearchFactory = $this->getVariationSearchFactory();
         $result = $itemSearchService->getResults([
-            $variationSearchFactory
+            $this->getVariationSearchFactory()
+                ->hasItemId($productId)
                 ->withVariationProperties()
                 ->withPrices()
                 ->isVisibleForClient()
                 ->isActive()
                 ->withResultFields($resultFields)
-                ->hasItemId($productId)
         ]);
 
         $this->logger->error('result', $result);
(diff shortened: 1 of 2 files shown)


TobiasGraml11 (approved):

src/Services/SearchService.php
```diff
@@ -465,7 +483,7 @@ private function getVariationIdForRedirect(string $query, array $documents)
         foreach ($documents as $document) {
             $variation = $document['data']['variation'];
             $barcodes = $document['data']['barcodes'] ?? [];
-            $variationPrice = $this->getCheapestPrice($document['data']['salesPrices']);
+            $variationPrice = $document['data']['prices']['default']['price']['value'] ?: 0;
```
TheKeymaster, line 486: Is this actually the cheapest price?

src/Services/SearchService.php
```diff
@@ -410,26 +418,36 @@ private function getProductDetailUrl(Response $response)
             $productId = explode('_', $productId)[0];
         }
 
-        $result = $itemSearchService->getResult(
-            $this->getSearchFactory()->hasItemId($productId)
-        );
+        // Necessary to retrieve variation data and prices
+        $resultFields = $this->loadResultFieldTemplate(ResultFieldTemplate::TEMPLATE_LIST_ITEM);
```
TheKeymaster, line 422: How do you find this? Did you check the Ceres code for this?

tests/Services/SearchServiceTest.php
```diff
@@ -192,27 +193,52 @@ public function testItemVariantIdExtractingForRedirectUrlGeneration()
...
+                'total' => 1,
+                'documents' => $this->getMultipleItemsDocuments(
+                    [
+                            '0' => [
+                                'id' => $mainVariationId,
+                                'price' => 10.00,
+                                'isMain' => true
+                            ]
+                        ]
+                )
```
TheKeymaster, line 233: Please check the indentation.

TheKeymaster (changes requested):

src/Services/SearchService.php
dblaichinger: Note: Also check why unit test isn't failing.

src/Services/SearchService.php
dblaichinger: Discussed in call

src/Services/SearchService.php
dblaichinger: > Is this actually the cheapest price?

Yes, the cheapeast price gets set on line 518, so the logic works as before, just without need to parse the `salesPrices` array.

TheKeymaster (approved):

PR #313, opened 2022-04-12
PLENTY-438_SSR_cat_filter

dblaichinger:


dblaichinger:
Issue is fixed with #315

PR #314, opened 2022-05-09
PLENTY-439 Ensure PHP8 compatibility

TobiasGraml11:


composer.json
```diff
@@ -3,13 +3,13 @@
   "type": "project",
   "description": "Findologic api plugin for plentymarkets system",
   "require": {
-    "php": "^7.0",
-    "findologic/http_request2": "2.3.1",
+    "php": "^8.0",
     "ext-simplexml": "*",
-    "squizlabs/php_codesniffer": "^3.5"
+    "findologic/http_request2": "2.4",
```
dblaichinger, line 8: https://github.com/findologic/HTTP_Request2/releases

> Establish support for PHP 7.4.

Can you please verify if 2.4. works with php8 as well.

tests/Api/Request/RequestBuilderTest.php
```diff
@@ -168,13 +168,13 @@ public function buildProvider()
     /**
      * @dataProvider buildProvider
      *
-     * @param string $userIp
+     * @param string|bool $userIp
```
dblaichinger, line 171: Why is this a boolean when "No user ip provided"? Should be empty string.

tests/Services/SearchServiceTest.php
```diff
@@ -150,8 +150,11 @@ public function testHandleSearchQuery()
 
         $itemSearchServiceMock = $this->getMockBuilder(ItemSearchService::class)
             ->disableOriginalConstructor()
-            ->setMethods([])
             ->getMock();
+        $itemSearchServiceMock->expects($this->once())
+            ->method('getResult')
+            ->willReturn(['success' => false, 'total' => 0]);
```
dblaichinger: Returning 'success' shouldn't be necessary anymore, plenty doesn't provide this anymore in the responses of the new search factories.

tests/Services/SearchServiceTest.php
```diff
@@ -194,8 +197,10 @@ public function testItemVariantIdExtractingForRedirectUrlGeneration()
...
```
dblaichinger: Returning 'success' shouldn't be necessary anymore

dblaichinger (changes requested):

tests/Services/SearchServiceTest.php
TobiasGraml11: Why do we check it in the code then?
https://github.com/findologic/plugin-plentymarkets-ceres-api/blob/0674c6a6d9b2d3d9817b232fd42ca9a9d82eac40/src/Services/SearchService.php#L354

tests/Api/Request/RequestBuilderTest.php
TobiasGraml11: Because `false` is returned by `RequestBuilder::getUserIp`, when no IP exists

tests/Services/SearchServiceTest.php
dblaichinger: This is an old version, please check develop: https://github.com/findologic/plugin-plentymarkets-ceres-api/blob/develop/src/Services/SearchService.php#L359

tests/Api/Request/RequestBuilderTest.php
dblaichinger: Ok it's used as bool check as well, not the best implementation but we'll leave it for now.

tests/Services/SearchServiceTest.php
TobiasGraml11: Why is the `ssr_update_deps` branch not updated with develop changes?

dblaichinger: That's where I'm currently on: https://github.com/findologic/plugin-plentymarkets-ceres-api/pull/315
Ready to be reviewed soon.

TobiasGraml11: If I would change this here now, the tests would fail. So I'll stick with it, until the code was also updated.

composer.json
TobiasGraml11: I checked the code in the IDE and ran the tests with PHP8.0.
Seems to have no issues.

tests/Services/SearchServiceTest.php
TobiasGraml11: Same as above

dblaichinger (approved):

PR #334, opened 2022-09-20
PLENTY-455 No alivetest for shop pages

TobiasGraml11:


src/Middlewares/Middleware.php
```diff
@@ -69,13 +70,22 @@ public function before(Request $request)
...
             return;
         }
 
         $this->eventDispatcher->listen(
             'IO.Resources.Import',
             function (ResourceContainer $container) {
+                /** @var CategoryService $categoryService */
+                $categoryService = pluginApp(CategoryService::class);
+                $isCategoryPage = $categoryService->getCurrentCategory() !== null;
+                $isInSearchOrCategoryPage = $this->isSearchPage || $isCategoryPage;
```
TobiasGraml11: - We should probably initialize `$this->isSearchPage` before that, just to be sure
- We should check for `$this->activeOnCatPage` as well

src/Middlewares/Middleware.php
```diff
@@ -88,7 +98,7 @@ function (ResourceContainer $container) {
                         'isSearchPage' => $this->isSearchPage,
                         'activeOnCatPage' => $this->activeOnCatPage,
                         'minimalSearchTermLength' => $this->pluginConfig->getMinimalSearchTermLength(),
-                        'languagePath' => $this->getLanguagePath(),
```
TobiasGraml11: Please keep that comma

src/Middlewares/Middleware.php
```diff
@@ -53,7 +54,7 @@ class Middleware extends PlentyMiddleware
     public function __construct(
         PluginConfig $pluginConfig,
         SearchService $searchService,
-        Dispatcher $eventDispatcher
+        Dispatcher $eventDispatcher,
```
TobiasGraml11: Please remove that comma, as this would break for shops still not on PHP8

PR #348, opened 2022-11-25
PLENTY-444 migrating remaining components

TobiasGraml11:
Follow up to #327


Mariusslo:
@TobiasGraml11 can you review this, thanks!

resources/views/Category/Item/Partials/Promotion.twig
```diff
@@ -1,7 +1,6 @@
...
-            <img class="image" src="{{ promotion.image }}" />
-        </a>
-    </div>
-{% endif %}
+<div>
+    <promotion-component
+            v-if="{{ promotion | length}} > 0"
+            :promotion="{{ promotion | json_encode}}"
+    />
+</div>
```
TobiasGraml11, line 6: Please make sure to have a newline on every file

resources/views/Category/Item/Partials/Promotion.twig
```diff
@@ -1,7 +1,6 @@
-{% if promotion is not empty and promotion.image is not empty %}
-    <div id="fl-promotion" class="page-content container-max">
-        <a href="{{ promotion.link }}">
-            <img class="image" src="{{ promotion.image }}" />
-        </a>
-    </div>
-{% endif %}
+<div>
```
TobiasGraml11, line 1: Is this `<div>` wrapper really needed?

resources/js/src/app/components/SearchTestComponent.vue
```diff
@@ -0,0 +1,23 @@
...
+    }
+  }
+};
+</script>
+
+<style scoped lang="scss">
+.fl-red-text {
+  color: red;
+}
+</style>
```
TobiasGraml11: I guess this test component is not needed anymore

resources/views/Category/Item/Partials/SmartDidYouMean.twig
```diff
@@ -1,8 +1,6 @@
 <div class="row mt-3">
```
TobiasGraml11, line 1: Can we add this div to the Vue component?

resources/views/Category/Item/Partials/SmartDidYouMean.twig
```diff
@@ -1,8 +1,6 @@
...
-    <div class="col-12">
-        <h1 class="h2" id="searchPageTitle">{{ query_info_message | striptags('<strong>,<a>') | raw }}</h1>
-        {% if smart_did_you_mean and smart_did_you_mean|length > 0 %}
-            <p class="smart-did-you-mean">{{ smart_did_you_mean | striptags('<strong>,<a>') | raw }}</p>
-        {% endif %}
-    </div>
+    <smart-did-you-mean
+            :smart-text="{{ smart_did_you_mean | json_encode}}"
+            :query-info-message="{{  query_info_message | striptags('<strong>,<a>') | raw | json_encode }}"
```
TobiasGraml11, line 4: By the look of it, this prop values are strings including HTML. I don't think we should run `| json_encode` on it.

resources/js/src/app/components/SmartDidYouMean.vue
```diff
@@ -0,0 +1,34 @@
...
+      type: String,
+      default: '',
+    },
+  }
+});
+</script>
+
+<style scoped>
+
+</style>
```
TobiasGraml11: Empty style blocks can be removed

resources/views/Category/Item/Partials/SearchFilters.twig
```diff
@@ -1,12 +1,9 @@
-{{ component( "Findologic::ItemList.Components.Filter.ItemFilterList") }}
-
 <div class="list-controls">
```
TobiasGraml11: Didn't we discuss, that this can be removed now? (Including the container link in the plugin.json)

resources/views/content/scripts.twig
```diff
@@ -37,40 +37,14 @@
...
-                                    v-bind:fallback-image-image-filter="'{{ plugin_path("Findologic") }}/images/no-picture-2.png'">
-                                </findologic-item-filter>
-                            <!-- /SSR -->
-                        </div>
-                    </client-only>
-                </div>
-            </div>
-        </div>
-    </div>
+    <findologic-filter-wrapper
```
TobiasGraml11, line 40: It it not required to add it like in the `item-filter-tag-list`?
`{{ component( "Findologic::ItemList.Components.Filter.FindologicItemFilterTagList" ) }}`

If not, we can probably also remove it for the tag list.

resources/views/ItemList/Components/Filter/ItemFilterList.twig
```diff
@@ -1,75 +0,0 @@
...
-                                :show-category-filter="showCategoryFilter"
-                                :facet="facet"
-                                :key="facet.id"
-                            ></findologic-item-filter>
-                        </div>
-                    </client-only>
-                </div>
-            </div>
-        </div>
-    </div>
```
TobiasGraml11, line 73: The filters are currently not split up into main/other filters. We will need to copy this part to the FindologicFilterWrapper.vue and make sure the other filters are expandable.

Have a look at the ItemFilterList.vue component.

resources/js/src/app/components/itemList/filter/ItemFilterList.vue
```diff
@@ -1,3 +1,77 @@
+<template>
```
TobiasGraml11: This should not be needed anymore, after removing the container link. The FindologicFilterWrapper should be the replacement.

resources/js/src/app/components/itemList/filter/FindologicFilterWrapper.vue
```diff
@@ -0,0 +1,95 @@
...
+        <div
+          class="container-max component-loading page-content"
+          :class="{ isLoading: isLoading }"
+        >
+          <div class="card-columns row">
+            <client-only>
+              <div class="w-100">
+                <!-- SSR:template(findologic-item-filter) -->
+                <findologic-item-filter
+                  template-override="#vue-findologic-item-filter"
```
TobiasGraml11: Should not be needed anymore
```suggestion
```

resources/js/src/app/components/itemList/filter/FindologicFilterWrapper.vue
```diff
@@ -0,0 +1,95 @@
...
+                  :facet="facet"
+                  :key="facet.id"
+                  :filtersPerRow="filtersPerRow"
+                  :fallbackImageColorFilter="fallbackImageColorFilter"
+                  :fallbackImageImageFilter="fallbackImageImageFilter"
+                  :current-category="currentCategory"
+                  :show-category-filter="showCategoryFilter"
+                />
+
+                <!-- /SSR -->
```
TobiasGraml11: I think the SSR:template comment is only required with separate .twig and .js files. We should be able to remove it.

```suggestion
                <findologic-item-filter
                  template-override="#vue-findologic-item-filter"
                  v-for="facet in facets"
                  :facet="facet"
                  :key="facet.id"
                  :filtersPerRow="filtersPerRow"
                  :fallbackImageColorFilter="fallbackImageColorFilter"
                  :fallbackImageImageFilter="fallbackImageImageFilter"
                  :current-category="currentCategory"
                  :show-category-filter="showCategoryFilter"
                />
```

resources/js/src/app/components/itemList/filter/ItemRangeSlider.vue
```diff
@@ -58,7 +58,8 @@ import { FacetAware, TemplateOverridable } from '../../../shared/interfaces';
 import { computed, defineComponent, onMounted, ref, watch } from '@vue/composition-api';
 import UrlBuilder, { PriceFacetValue } from '../../../shared/UrlBuilder';
 import TranslationService from '../../../shared/TranslationService';
-import * as noUiSlider from 'noUiSlider';
+import * as noUiSlider from 'nouislider';
+import 'nouislider/dist/nouislider.css';
```
TobiasGraml11: Would it be simpler to include it within the `<style>` block? Maybe the test adaptions for the styling would not be needed anymore.

resources/js/src/app/components/itemList/filter/ItemRangeSlider.vue
```diff
@@ -179,9 +180,9 @@ export default defineComponent({
           }
         });
 
-        slider.on('update', function (ui: string[]) {
-          valueFrom.value = ui[0];
-          valueTo.value = ui[1];
+        slider.on('update', function (values: (number | string)[]) {
+          valueFrom.value = values[0];
+          valueTo.value = values[1];
```
TobiasGraml11: As this was only `string[]` before, do we need to parse the numbers to strings now?

src/Middlewares/Middleware.php
```diff
@@ -93,14 +93,17 @@ function (ResourceContainer $container) {
                 $categoryService = pluginApp(CategoryService::class);
                 $isCategoryPage = $categoryService->getCurrentCategory() !== null && $this->activeOnCatPage;
                 $isInSearchOrCategoryPage = $this->isSearchPage || $isCategoryPage;
+                $currentCategory = $categoryService->getCurrentCategory();
```
TobiasGraml11, line 96: Please define `$currentCategory` some lines above, so it can be re-used for `$isCategoryPage`

src/Middlewares/Middleware.php
```diff
@@ -93,14 +93,17 @@ function (ResourceContainer $container) {
...
-                if ($isInSearchOrCategoryPage && !$this->searchService->aliveTest()) {
-                    return false;
+                $showCategoryFilter = true;
+
+                // Show category filter only in the 0 and 1 level categories
+                if ($currentCategory !== null && $currentCategory->level > 1) {
+                    $showCategoryFilter = false;
                 }
 
-                if ($this->pluginConfig->get(Plugin::CONFIG_LOAD_NO_UI_SLIDER_STYLES_ENABLED)) {
```
TobiasGraml11, line 101: Some points here:
- I forgot that this configuration exists, so we will probably need to check it, when importing the CSS within the components. But I am not sure, if that works that easily, as it is probably still ending up within the plugin build.
    - Not sure if that counts as hacky, but we can import it under a `fl-no-ui-slider` class and only add this class if the configuration is set accordingly: `.fl-no-ui-slider { @import 'no-ui-slider.css' }`
- Maybe it would make sense to move all of the configurations into one object, that can be forwarded as one prop. This way we could also create a type for it

resources/views/ItemList/Components/Filter/ItemFilterList.twig
```diff
@@ -1,75 +0,0 @@
-{# Include storefront template translations #}
-{% import "Ceres::PageDesign.Macros.Translations" as Translations %}
-{{ Translations.add( "Findologic", "Template" ) }}
```
TobiasGraml11, line 3: I currently do not see any translations, this will still need to be loaded somewhere:

image

resources/views/Category/Item/Partials/Promotion.twig
Mariusslo: I am not sure, this change was not made by me it probably came with the older commits. Where should I investigate on the frontend?

resources/js/src/app/components/itemList/filter/ItemRangeSlider.vue
Mariusslo: There are a lot of styles in that css i think this is better for readability, also the mock for the css files don't make it that more complex IMO

resources/js/src/app/components/itemList/filter/ItemRangeSlider.vue
Mariusslo: With the current config probably not, could add a check if needed?

resources/views/Category/Item/Partials/SmartDidYouMean.twig
Mariusslo: I am not sure about this, do you know where to look for it on the front end?

resources/views/Category/Item/Partials/Promotion.twig
TobiasGraml11: You can add promotions this way: https://docs.findologic.com/doku.php?id=integration_documentation:plugin:en:api:plentymarkets:promotions#using_promotions_with_plentymarkets_ceres

resources/views/Category/Item/Partials/SmartDidYouMean.twig
TobiasGraml11: Check this out: https://docs.findologic.com/doku.php?id=integration_documentation:plugin:en:api:plentymarkets:smart_did_you_mean

resources/js/src/app/components/itemList/filter/ItemRangeSlider.vue
TobiasGraml11: Let's just use `values[0].toString()`, the function exists for numbers and strings

resources/js/src/app/components/itemList/filter/ItemRangeSlider.vue
TobiasGraml11: Maybe I need to clarify, that I don't want you to copy the file content, just something like this:
```
<style scoped lang="scss">
@import 'nouislider/dist/nouislider.css'
</style>
```
I think it helps the readability, as the code and the styles are clearly separated.

And maybe it's even needed, with my suggestion from https://github.com/findologic/plugin-plentymarkets-ceres-api/pull/348#discussion_r1082461662. Except we find another solution for that.

resources/views/Category/Item/Partials/Promotion.twig
Mariusslo: Apparently, without the div, it breaks the page, showing no filters or content. Don't know why

TobiasGraml11: Any console logs?

resources/views/Category/Item/Partials/SmartDidYouMean.twig
Mariusslo: Doesn't work without json_encode

resources/js/src/app/components/itemList/filter/ItemFilterList.vue
Mariusslo: So you mean all of its content should be in the wrapper component?

resources/views/Category/Item/Partials/Promotion.twig
Mariusslo: No

Mariusslo: Should I leave the divs or find another solution?

TobiasGraml11: Let's just leave them there, if they break otherwise

resources/js/src/app/components/itemList/filter/ItemFilterList.vue
TobiasGraml11: Yes, right know both exist with some duplicated code.

FindologicFilterWrapper should be the component for all use cases. (Which is probably only one, after removing the container link)

resources/views/Category/Item/Partials/SmartDidYouMean.twig
TobiasGraml11: Weird, then we will need to leave it there

Mariusslo:
@TobiasGraml11, can you check the failing jslint test? I am not sure why that's happening

TobiasGraml11:
@Mariusslo Please check out my commits.
I guess we can cleanup the dependencies a bit more. By using vite, `@vue/cli-service` should not be needed anymore. Just one example.

package.json
```diff
@@ -6,74 +6,52 @@
...
   "dependencies": {
     "@tanem/svg-injector": "^10.1.19",
+    "@vue/cli-plugin-eslint": "^5.0.8",
     "@vue/composition-api": "^1.0.0-rc.5",
-    "vue": "^2.6.11",
+    "vue": "^2.6.14",
     "vue-class-component": "^7.2.3",
-    "vue-property-decorator": "^9.1.2"
+    "vue-property-decorator": "^9.1.2",
+    "vue-template-compiler": "2.6.14"
```
TobiasGraml11: Any reason why `vue-template-compiler` was moved from the `devDependencies` and changed from `^2.6.14` to `2.6.14`?

package.json
```diff
@@ -6,74 +6,52 @@
...
-    "babel-plugin-transform-object-rest-spread": "^6.26.0",
-    "babel-polyfill": "^6.26.0",
-    "babel-preset-es2015": "^6.24.1",
-    "css-loader": "^5.0.2",
-    "eslint": "^8.10.0",
-    "eslint-plugin-vue": "^8.5.0",
+    "eslint": "^8.35.0",
+    "eslint-plugin-vue": "^8.7.1",
     "jest": "^27.5.1",
     "jquery": "^3.6.0",
```
TobiasGraml11, line 42: Please check if `jquery` is still used. I don't think that's the case.

package.json
```diff
@@ -6,74 +6,52 @@
...
-    "node-sass": "^7.0.1",
     "nouislider": "^15.5.1",
-    "postcss-loader": "^6.2.1",
     "sass": "^1.49.9",
-    "sass-loader": "^12.6.0",
-    "style-loader": "^3.3.1",
     "terser-webpack-plugin": "^5.3.1",
```
TobiasGraml11: Webpack was removed, is `terser-webpack-plugin` still needed?

package.json
TobiasGraml11: This also includes `@types/jquery` and `@types/jqueryui`

package.json
```diff
@@ -6,74 +6,52 @@
...
-    "@vue/cli-plugin-eslint": "^5.0.1",
-    "@vue/cli-plugin-typescript": "^5.0.1",
-    "@vue/cli-plugin-unit-jest": "^5.0.1",
-    "@vue/cli-service": "^5.0.1",
+    "@vitejs/plugin-vue": "^1.6.1",
     "@vue/eslint-config-typescript": "^10.0.0",
     "@vue/test-utils": "^1.3.0",
-    "@vue/vue2-jest": "^27.0.0-alpha.4",
+    "@vue/vue2-jest": "^27.0.0",
     "autoprefixer": "^9.0.0",
```
TobiasGraml11: `autoprefixer` is probably also not needed anymore.

resources/js/src/app/modules/filter.ts
```diff
@@ -10,7 +10,10 @@ import ItemRangeSlider from '../components/itemList/filter/ItemRangeSlider.vue';
 import FindologicItemFilter from '../components/itemList/filter/FindologicItemFilter.vue';
 import FindologicItemFilterTagList from '../components/itemList/filter/FindologicItemFilterTagList.vue';
 import ItemFilterImage from '../components/itemList/filter/ItemFilterImage.vue';
-import ItemFilterList from '../components/itemList/filter/ItemFilterList.vue';
+import SmartDidYouMean from '../components/SmartDidYouMean.vue';
+import PromotionComponent from '../components/PromotionComponent.vue';
+import FindologicFilterWrapper from '../components/itemList/filter/FindologicFilterWrapper.vue';
+
```
TobiasGraml11, line 16:
```suggestion
```

resources/views/content/scripts.twig
```diff
@@ -37,71 +37,47 @@
...
-                               @keyup.down="keydown()"
-                               @keyup.up="keyup()"
-                               @focus="isSearchFocused = true"
-                               @blur="onBlurSearchField($event)">
+                               v-model="searchString" 
+                               @keyup.enter="search()" 
+                               @input="onValueChanged($event.target.value)"
+                               @focus="isSearchFocused = true" 
+                               @blur="onBlurSearchField($event)" 
+                               :autofocus="isShopBuilder"
```
TobiasGraml11: Please remove the autofocus, as our suggest opens up on focus.
We already had issues here, where it opened on every page.

resources/views/content/translations.twig
```diff
@@ -0,0 +1,2 @@
+{% import "Ceres::PageDesign.Macros.Translations" as Translations %}
+{{ Translations.add( "Findologic", "Template" ) }}
```
TobiasGraml11: Please add a newline

src/Providers/FindologicServiceProvider.php
```diff
@@ -19,7 +22,15 @@ public function register()
         $this->getApplication()->singleton(SearchService::class);
         $this->getApplication()->singleton(PluginConfig::class);
         $this->getApplication()->singleton(PluginConfigurationValidator::class);
+        // $this->getApplication()->register(FindologicTemplateProvider::class);
```
TobiasGraml11: Please remove this line, or uncomment it if it is needed

vite.config.js
```diff
@@ -0,0 +1,22 @@
...
+                app: './resources/js/src/index.ts',
+            },
+            output: {
+                dir: './resources/js/dist/',
+                entryFileNames: 'findologic_ceres.js',
+                assetFileNames: 'findologic_ceres.css',
+            }
+        },
+    },
+})
```
TobiasGraml11: Please add a newline

package.json
Mariusslo: 1) moved - no reason, readded to dev
2) version changed because it is required that Vue and template-compiler would be exactly the same version

Note: when migrating to vue 2.7 I believe the template compiler will be completely removed

package.json
Mariusslo: jQuery is still used in at least 1 place, should we refactor it?

https://github.com/findologic/plugin-plentymarkets-ceres-api/blob/PLENTY-444-migrating_remaining_components/resources/js/src/app/components/itemList/filter/ItemRangeSlider.vue#L163

TobiasGraml11: Seems like it is using the jquery of the shop, and not the package, as the package would need to be imported somewhere.

Mariusslo: It is also used in tests
image

TobiasGraml11: Ok, then let's leave it for now

PR #350, opened 2022-12-12
PLENTY-458 price filter range

Mariusslo:


resources/js/src/app/components/itemList/filter/ItemRangeSlider.vue
```diff
@@ -144,6 +144,21 @@ export default defineComponent({
...
+      // Determine number of decimals in the slider
+      let decimalNumber = 2;
+
+      if(props.facet.step === 1) {
+        decimalNumber = 0;
+      }
+
+      if(props.facet.step === 0.1) {
+        decimalNumber = 1;
+      }
```
TobiasGraml11, line 160: How about setting the configuration value already to the correct number?
So use the current caption as value as well.

resources/lang/de/Config.properties
```diff
@@ -19,3 +19,4 @@ loadNoUiSliderStylesEnabledLabel=no-ui-slider Styles laden (für Filtertyp Berei
 loadNoUiSliderStylesDisabled=Nein
 loadNoUiSliderStylesEnabled=Ja
 minimalSearchTermLengthLabel=Minimale Suchwortlänge
+priceRangeFilterSteps=Dezimalpunkte im Preisbereichsfilter
```
TobiasGraml11: "No newline at end of file"

resources/lang/en/Config.properties
```diff
@@ -19,3 +19,4 @@ loadNoUiSliderStylesEnabledLabel=Load no-ui-slider styles (for filter type range
 loadNoUiSliderStylesDisabled=No
 loadNoUiSliderStylesEnabled=Yes
 minimalSearchTermLengthLabel=Minimal search term length
+priceRangeFilterSteps=Decimal points in price range filter
```
TobiasGraml11: "No newline at end of file"

src/Api/Response/Parser/FiltersParser.php
```diff
@@ -2,17 +2,18 @@
...
 use Plenty\Modules\Plugin\Libs\Contracts\LibraryCallContract;
-use SimpleXMLElement;
 
 /**
  * Class FiltersParser
  * @package Findologic\Api\Response\Parser
  */
 class FiltersParser
-{
+{
```
TobiasGraml11:
```suggestion
{
```

src/Api/Response/Parser/FiltersParser.php
```diff
@@ -191,6 +196,12 @@ protected function parseFilter($filter, $isMainFilter = false)
             $filterData['step'] = (float)$filter->attributes->stepSize;
         }
 
+        if ($filterName === 'price' && $filterData['findologicFilterType'] === Plugin::FILTER_TYPE_RANGE_SLIDER) {
```
TobiasGraml11: Let's use this configuration for all range sliders.

```suggestion
        if ($filterData['findologicFilterType'] === Plugin::FILTER_TYPE_RANGE_SLIDER) {
```

tests/Api/Response/Parser/FiltersParserTest.php
```diff
@@ -50,7 +60,7 @@ public function testParse($response, array $expectedResult)
         $filtersParserMock = $this->getFiltersParserMock();
 
         $results = $filtersParserMock->parse(simplexml_load_string($response));
-
+
```
TobiasGraml11:
```suggestion

```

resources/lang/de/Config.properties
TobiasGraml11:
```suggestion
priceRangeFilterSteps=Dezimalstellen für Bereichsfilter
```

resources/lang/en/Config.properties
TobiasGraml11:
```suggestion
priceRangeFilterSteps=Decimal places for range sliders
```

TobiasGraml11 (changes requested):

resources/js/src/app/components/itemList/filter/ItemRangeSlider.vue
Mariusslo: Then we don't have the correct step variable to pass, as we need to configure values:
- for the step count (stepSize)
- for the decimals for formating

TobiasGraml11: Makes sense, thanks.

Then please use https://www.php.net/manual/de/control-structures.match.php instead of the if blocks.

TobiasGraml11: True, thanks!

package.json
```diff
@@ -18,6 +18,7 @@
   "dependencies": {
     "@tanem/svg-injector": "^10.1.19",
     "@vue/composition-api": "^1.0.0-rc.5",
+    "run": "^1.4.0",
```
TobiasGraml11: Why was this added now?

tests/Api/Response/Parser/FiltersParserTest.php
```diff
@@ -69,7 +78,8 @@ protected function getFiltersParserMock($methods = [])
 
         $filtersParserMock = $this->getMockBuilder(FiltersParser::class)
             ->setConstructorArgs([
-                'libraryCallContract' => $this->libraryCallContract
+                'libraryCallContract' => $this->libraryCallContract,
+                'configRepository' => $this->configRepository
```
TobiasGraml11, line 82: Just noticed, that we do not test if the configuration is set correctly. Could you add a test for that?

TobiasGraml11 (changes requested):

TobiasGraml11 (approved):

PR #360, opened 2023-04-13
PLENTY-335 Plenty SSR

Mariusslo:


resources/js/src/app/components/itemList/filter/FindologicFilterWrapper.vue
```diff
@@ -1,20 +1,20 @@
...
+        v-if="facets.some((e) => e.isMain === true)"
         class="ml-0 main-filters"
       >
         <div
           class="container-max component-loading page-content"
           :class="{ isLoading: isLoading }"
         >
           <div class="card-columns row">
-            <client-only>
+            <div>
```
TobiasGraml11: Can we just remove it, without replacing it with a `<div>`?

resources/js/src/app/components/itemList/filter/FindologicFilterWrapper.vue
```diff
@@ -49,22 +49,23 @@
...
           class="ml-0 filter-collapse collapse"
         >
           <div
             class="container-max component-loading page-content mb-5"
-            :class="{ 'isLoading': isLoading }"
+            :class="{ isLoading: isLoading }"
           >
             <div class="card-columns row">
-              <client-only>
+              <div>
```
TobiasGraml11: Can we just remove it, without replacing it with a `<div>`?

resources/js/src/app/components/itemList/filter/FindologicFilterWrapper.vue
```diff
@@ -1,20 +1,20 @@
 <template>
-  <client-only>
+  <div>
```
TobiasGraml11: Can we just remove it, without replacing it with a `<div>`?

resources/js/src/app/components/itemList/filter/FindologicFilterWrapper.vue
```diff
@@ -95,14 +96,14 @@ import {
   PlentyVuexStore,
   TemplateOverridable,
 } from '../../../shared/interfaces';
-import { computed, defineComponent } from '@vue/composition-api';
+import { computed, defineComponent, onMounted, ref } from '@vue/composition-api';
 import FindologicItemFilter from './FindologicItemFilter.vue';
 import type { PropType } from '@vue/composition-api';
 import TranslationService from '../../../shared/TranslationService';
 
 interface FindologicFilterWrapperProps extends TemplateOverridable, FacetAware {
```
TobiasGraml11, line 102: This does not seem to replicate all the props

resources/js/src/app/components/itemList/filter/FindologicItemFilterTagList.vue
```diff
@@ -1,32 +1,47 @@
 <template>
   <div class="selected-filters clearfix">
-    <span
+    <div
```
TobiasGraml11, line 3: Any specific reasoning for this change?

resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.vue
```diff
@@ -101,8 +99,13 @@ export default defineComponent({
   ],
 
   setup(props: CategoryDropdownProps, { root }) {
+    const pleaseSelectText = ref<string>('');
+    const categoryIsSelected = ref<boolean>(false);
+    const categories = ref<Array<FacetValue> | undefined>([]);
+      
     root.$options.template = props.template || '#vue-item-dropdown';
 
+
```
TobiasGraml11, line 108:
```suggestion
```

resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.vue
```diff
@@ -177,17 +180,21 @@ export default defineComponent({
 
     onMounted(() => {
       dropdownLabel.value = buildDropdownLabel() as string;
+      pleaseSelectText.value = TranslationService.translate('Findologic::Template.pleaseSelect');
+      categoryIsSelected.value = isSelected.value;
+      categories.value = comCategories.value;
     });
 
+
```
TobiasGraml11:
```suggestion
```

resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.vue
```diff
@@ -129,7 +132,7 @@ export default defineComponent({
       return typeof props.currentCategory !== 'undefined';
     });
 
-    const categories = computed((): FacetValue[] | undefined  => {
+    const comCategories = computed((): FacetValue[] | undefined  => {
```
TobiasGraml11, line 135: Why do we need to save the computed `comCategories` to a ref `categories` within onMounted? Can you use the computed within the template?

resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.vue
```diff
@@ -177,17 +180,21 @@ export default defineComponent({
 
     onMounted(() => {
       dropdownLabel.value = buildDropdownLabel() as string;
+      pleaseSelectText.value = TranslationService.translate('Findologic::Template.pleaseSelect');
+      categoryIsSelected.value = isSelected.value;
```
TobiasGraml11, line 182: Is `isSelected` even defined?

resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.vue
```diff
@@ -101,8 +99,13 @@ export default defineComponent({
   ],
 
   setup(props: CategoryDropdownProps, { root }) {
+    const pleaseSelectText = ref<string>('');
+    const categoryIsSelected = ref<boolean>(false);
```
TobiasGraml11, line 103: `categoryIsSelected` is never changed, so the variable is kind of useless. Can you just use `isCategorySelected`?

resources/js/src/app/components/itemList/filter/ItemRangeSlider.spec.ts
```diff
@@ -85,7 +88,7 @@ describe('ItemRangeSlider', () => {
         expect(UrlBuilder.updateSelectedFilters).toHaveBeenNthCalledWith(2, facet, 'price', { 'max': 140.55, 'min': 100 });
     });
 
-    // TODO: reenable the test once the tested functionality works properly
+    // // TODO: reenable the test once the tested functionality works properly
```
TobiasGraml11: I guess it can't be re-enabled already? Could you please try this test?

```suggestion
    // TODO: reenable the test once the tested functionality works properly
```

resources/js/src/app/components/itemList/filter/ItemRangeSlider.vue
```diff
@@ -144,9 +144,10 @@ export default defineComponent({
 
     onMounted(() => {
       const values = UrlBuilder.getSelectedFilterValue(props.facet.id);
+      TranslationService = TranslationServiceImport;
       valueFrom.value = (values ? values.min : props.facet.minValue) || '';
       valueTo.value = (values ? values.max : props.facet.maxValue) || '';
-
+      applyText.value = TranslationService.translate('Ceres::Template.itemApply');
```
TobiasGraml11, line 148: Can't you use the import directly?

TobiasGraml11 (changes requested):

resources/js/src/app/components/itemList/filter/FindologicItemFilterTagList.vue
Mariusslo: Yes, SSR recognized this html as faulty

resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.vue
Mariusslo: Wrong, `categoryIsSelected` is changed during onMounted method and references `isSelected = copmuted()` which uses DOM elements and can't be parsed during server side rendering

resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.vue
Mariusslo: yes, it's a computed property

resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.vue
Mariusslo: Because of the isCategorySelected method - it uses DOM elements, when we use the computed value it retrieves the categories during SSR and then tries to evaluate isCategorySelected which breaks everything, that is why we populate the data only onMounted, thus these 2 properties(computed and ref) come in play, it's kind of a hack at this point

TobiasGraml11: `isCategorySelected` is not used within the `comCategories` computed. How does it interfere here?

Mariusslo: in the html template we loop through categories and then call isCategorySelected on each

Mariusslo: further explanation. if we use comCategories in the template, the isCategorySelected is called during server rendering and breaks, if we use a ref in onMounted isCategorySelected is called during mounting

resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.vue
TobiasGraml11: Ah, I've overseen that one

resources/js/src/app/components/itemList/filter/ItemCategoryDropdown.vue
TobiasGraml11: Ah ok :+1:

PR #369, opened 2023-09-04
PLENTY-480 It should be possible to use the range slider component multiple times

aserstobitov:


TobiasGraml11 (approved):

resources/js/src/app/components/itemList/filter/ItemRangeSlider.vue
```diff
@@ -82,7 +82,8 @@ export default defineComponent({
       return 'fl-range-slider-' + props.facet.id
           .replace(/\W/g, '-')
           .replace(/-+/, '-')
-          .replace(/-$/, '');
+          .replace(/-$/, '')
+          + Math.random().toString(36).substr(2, 9);
```
TobiasGraml11: I am a bit worried about the random string within the `computed` property. Because by the nature of `computed`, it will update the property every tick.
It would maybe be saver to set this random string as a constant at the beginning of the setup and just use it in this line.

TobiasGraml11 (changes requested):

TobiasGraml11 (approved):

PR #371, opened 2023-10-02
Plenty 480 use range slider multiple times SSR version fix

aserstobitov:


resources/js/src/app/components/itemList/filter/ItemRangeSlider.vue
```diff
@@ -28,7 +28,7 @@
     <div class="row fl-range-slider-row">
       <div class="col-md-9 col-sm-9 col-xs-12">
         <div
-          :id="sanitizedFacetId"
```
TobiasGraml11: Please keep the quotes

TobiasGraml11 (changes requested):

resources/js/src/app/components/itemList/filter/ItemRangeSlider.vue
aserstobitov: done, thanks for noticing

TobiasGraml11 (changes requested):
There is also a merge conflict for the build file. Could you please run another build?
If it's not resolved after that, please manually resolve it on the PR.

TobiasGraml11 (approved):

PR #385, opened 2024-05-14
Plenty 472 migrate response to json with variants without api

rimvydas-hashtages:


tests/MockResponses/someResultsWithFilters.json
TobiasGraml11, line 1: - Guess we can remove the corresponding XML file?
- What about noResults.xml, is it not used or why don't we need a JSON version of it?

src/Api/Response/ResponseParser.php
```diff
@@ -51,171 +53,121 @@ public function parse(HttpRequest $request, $responseData): Response
...
 
-                foreach ($product->properties->property as $property) {
-                    $productData['properties'][$property['name']->__toString()] = $property->__toString();
-                }
+        $queryStringType = null;
 
-                $products[] = $productData;
-            }
+        if($improvedQuery){
+            $queryStringType = 'improved';
```
TobiasGraml11: 1. The response class has constants for the query types
2. Do we never need the `didYouMean` type?

src/Api/Response/Parser/FiltersParser.php
```diff
@@ -107,45 +105,49 @@ public function parseForWidgets($data): array
...
-    public function parseFilterItem($filterType, &$filterItem, $data, $index)
+    public function parseFilterItem($filterType, &$filterItem, array $data, $index)
     {
         if (!empty($data)) {
             $filterItem['items'] = [];
-            $filterItem['name'] = $data->name->__toString();
-            $filterItem['position'] = $index;
-            $filterItem['count'] = $data->frequency->__toString();
+            $filterItem['name'] = $data['value'];
+            $filterItem['position'] = "item";
```
TobiasGraml11: Why did you change the `position` value?

src/Api/Response/Parser/FiltersParser.php
```diff
@@ -107,45 +105,49 @@ public function parseForWidgets($data): array
...
+            $filterItem['count'] = array_key_exists('frequency', $data) ? (string) $data['frequency'] : '';
```
TobiasGraml11: `$data['frequency] ?? ''` should be enough. The operator has the key check "included". Can probably be applied in some other places as well.

TobiasGraml11 (changes requested):

src/Api/Response/Parser/FiltersParser.php
rimvydas-hashtages: @TobiasGraml11 why I named it 'item' everywhere is because in the develop branch every filter item basically has 'item' in it as the position
image

TobiasGraml11: Does the `$index` variable contain "item"? If so, I think it still makes sense to keep it like it is and don't hardcode it there.

rimvydas-hashtages: No, it contains numbers. Guess in xml it was always 'item' cause the xml is something like this:
`<items>
   <item>
      <id>....
    </item>
 </items>`

TobiasGraml11: Interesting, please still stick with `$filterItem['position'] = $index;`. I don't think it's used anywhere, but having the numeric values makes more sense.

rimvydas-hashtages: @TobiasGraml11 check it again now

src/Api/Response/ResponseParser.php
TobiasGraml11: @rimvydas-hashtages Only one more thing: Please let me know why we never need to set the `didYouMean` type.

rimvydas-hashtages: Because old code always overrides the type to 'did-you-mean' whenever the 'didYouMeanQuery' is set:
image
image

TobiasGraml11 (approved):

